PR #4, opened 2020-12-14
Provides Initial Project Structure

sumit-nathany:
This PR intends to create the basic folder structure to be used by the rSocket Swift Library. This is in no way a final structure and developers are welcome to provide feedback.

### DIFF
--- a/Example/Podfile
+++ b/Example/Podfile
@@ -1,0 +1,14 @@
+# Platform version to build against
+platform :ios, '11.0'
+
+use_frameworks!
+
+target 'RSocketSwift_Example' do
+  pod 'RSocketSwift', :path => '../'
+
+  target 'RSocketSwift_Tests' do
+    inherit! :search_paths
+
+    
+  end
+end
--- a/.gitignore
+++ b/.gitignore
@@ -1,18 +1,7 @@
 # Xcode
+.DS_Store
 #
-# gitignore contributors: remember to update Global/Xcode.gitignore, Objective-C.gitignore & Swift.gitignore
-
-## User settings
-xcuserdata/
-
-## compatibility with Xcode 8 and earlier (ignoring not required starting Xcode 9)
-*.xcscmblueprint
-*.xccheckout
-
-## compatibility with Xcode 3 and earlier (ignoring not required starting Xcode 4)
 build/
-DerivedData/
-*.moved-aside
 *.pbxuser
 !default.pbxuser
 *.mode1v3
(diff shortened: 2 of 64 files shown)


OlegDokuka (changes requested):
Looks like the project structure proposed there is a default xcode for iOS project which I'm not sure is something we wanna have for a generic swift library. 

Assume we want to use the same code for iOS | desktop | potentially server side

sumit-nathany:
So, theoretically, Swift libraries don’t need Xcode. But for all practical purposes, Swift libraries are implemented to work with Apple systems only (iOS/macOS/tvOS/watchOS) unless otherwise specified. Moreover, Swift is still a relatively newer language and even though powerful, it doesn’t contain many important features — e.g. - Multithreading. For multi threading, library owners rely on Apple-provided implementations, e.g. - GCD. So, they need to target it to Apple systems.
This usage pattern evolves with time. Swift, unlike Java or most other languages, is majorly used today with Apple systems with only a handful of developers using it with other systems. Most of the usage outside Apple ecosystem is for exploratory and academic purposes.
With that being said, we are open to discuss whether we want this to be a pure Swift library which is decoupled with underlying OSes or we want to target this to Apple systems (at least initially). As of now, the library DOESN’T contain any Apple OS related code (in fact, it doesn’t contain any code at all in this PR :slightly_smiling_face:) and so it can be compiled with Swift rather than `xcodebuild`. But going forward, that may change.
The unnecessary folders like assets are part of the Example app which makes use of this library. This is also a standard pattern used in the community. Again, this behaviour can change depending on the outcome of our discussion whether we want this to a be a pure Swift library or not.

Example/Podfile
```diff
@@ -0,0 +1,14 @@
...
+
+target 'RSocketSwift_Example' do
+  pod 'RSocketSwift', :path => '../'
+
+  target 'RSocketSwift_Tests' do
+    inherit! :search_paths
+
+    
+  end
+end
```
nkristek, line 14: Just a minor suggestion, once we have SwiftPM support let's base the example project on SwiftPM instead of Cocoapods, I can speak from experience that I don't even have Cocoapods installed so I wouldn't be able to run the example project.

nkristek (commented):
How about we name the library `RSocket` instead of `RSocketSwift`?

nkristek:
I've put up a draft PR to start a discussion around a multi library solution: #5

sumit-nathany:
Moving these discussions to #5

PR #5, opened 2020-12-15
adds initial project structure

nkristek:
This PR shows my suggestion for a project structure which supports the multi library approach we talked about. As seen in the `Package.swift` file there are 3 libraries:
- **RSocketCore**: Has no dependencies and contains only parsing logic. Maybe it needs `SwiftNIO` for `ByteBuffer` (depending on which path we want to take regarding `Data` vs. `ByteBuffer`)
- **RSocketCombine**: Has a dependency on `RSocketCore` to access the parsing logic and implements `Combine` publishers
- **RSocketReactiveSwift**: Has a dependency on `RSocketCore` to access the parsing logic and `ReactiveSwift` to implement it using `SignalProducer`

The example project is not part of this PR as we should focus on the main library first.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,0 +1,16 @@
+{
+  "object": {
+    "pins": [
+      {
+        "package": "ReactiveSwift",
+        "repositoryURL": "https://github.com/ReactiveCocoa/ReactiveSwift.git",
+        "state": {
+          "branch": null,
+          "revision": "e03cda84105ba707de039b757e2b2de868c65a3e",
+          "version": "6.5.0"
+        }
+      }
+    ]
+  },
+  "version": 1
+}
--- a/Package.swift
+++ b/Package.swift
@@ -1,0 +1,30 @@
+// swift-tools-version:5.3
+
+import PackageDescription
+
+let package = Package(
+    name: "RSocket",
+    platforms: [
+        .macOS(.v10_15),
+        .iOS(.v9),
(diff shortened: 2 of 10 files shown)


OlegDokuka:
I like this approach.

I would suggest keeping naming the same between all the projects, hence I would suggest renaming RSocket to RSocketCore.

Apart, so far soo good

Marcocanc:
I like this as well!

ciychodianda:
I like this approach. This covers the core and API portions. How do we want to handle the transport?
Transports/RSocketSwiftNIO/RSocketSwiftNIO-Client/
OR
RSocket-Transport-SwiftNIO/RSocket-Transport-SwiftNIO-Client

OlegDokuka:
I'm for the `RSocket-Transport-SwiftNIO/RSocket-Transport-SwiftNIO-Client` to keep the convention between implementations. Or even `RSocket-Transport-SwiftNIO-Client` and `RSocket-Transport-SwiftNIO-Server`

OlegDokuka:
@nkristek moving PR to `Ready for review`. Please make sure your commits are signed to pass the DCO check

OlegDokuka (approved):
LGTM

ciychodianda (approved):
LGTM

Marcocanc (approved):

Samer-Abdulaziz (approved):
Looks good

PR #7, opened 2020-12-29
adds initial frame parsing logic

nkristek:
## Summary

This PR adds structs and en-/decoding logic required for serialisation of rSocket frames.

## Details

~~First and foremost, the parsing logic uses `Data` as its essential binary representation, as `Data` is contained in `Foundation` (standard library in Swift) and other binary representations like `ByteBuffer` are convertible to it.~~

~~To make working with `Data` a bit safer and remove boilerplate, I added a dependency on a very lightweight library called [BinaryKit](https://github.com/Cosmo/BinaryKit). It is a very efficient wrapper around `Data` which makes it easier to, for example, append single bits to a binary.
As it stands, we have created a [PR](https://github.com/Cosmo/BinaryKit/pull/22) to make parsing a bit more efficient for our use case.
If our goal is to not have such a dependency, we could replicate something similar which is optimised for our use case, but I leave that up for debate, as it also would be more code to maintain.~~

**Edit:** This PR now uses `ByteBuffer` for its serialisation.

I have structured the frame parsing for each frame in the following parts:
#### Frame struct:
<details>
  <summary>ErrorFrame.swift </summary>

```swift
public struct ErrorFrame {
    public let header: FrameHeader

    public let errorCode: ErrorCode

    public let errorData: String

    public init(
        header: FrameHeader,
        errorCode: ErrorCode,
        errorData: String
    ) {
        self.header = header
        self.errorCode = errorCode
        self.errorData = errorData
    }
}
```

</details>

#### Decoder:
<details>
  <summary>ErrorFrameDecoder.swift</summary>

```swift
public struct ErrorFrameDecoder: FrameDecoder {
    public func decode(header: FrameHeader, buffer: inout ByteBuffer) throws -> ErrorFrame {
        guard let codeValue: UInt32 = buffer.readInteger() else {
            throw FrameError.tooSmall
        }
        let errorCode = ErrorCode(rawValue: codeValue)
        let errorData: String
        if buffer.readableBytes > 0 {
            guard let string = buffer.readString(length: buffer.readableBytes) else {
                throw FrameError.stringContainsInvalidCharacters
            }
            errorData = string
        } else {
            errorData = ""
        }
        return ErrorFrame(
            header: header,
            errorCode: errorCode,
            errorData: errorData
        )
    }
}
```
  
</details>

#### Encoder:
<details>
  <summary>ErrorFrameEncoder.swift</summary>

```swift
public struct ErrorFrameEncoder: FrameEncoder {
    public func encode(frame: ErrorFrame, using allocator: ByteBufferAllocator) throws -> ByteBuffer {
        var buffer = try FrameHeaderEncoder().encode(header: frame.header, using: allocator)
        buffer.writeInteger(frame.errorCode.rawValue)
        if !frame.errorData.isEmpty {
            let bytesWritten = buffer.writeString(frame.errorData)
            guard bytesWritten > 0 else {
                throw FrameError.stringContainsInvalidCharacters
            }
        }
        return buffer
    }
}
```
  
</details>

I chose to have separate encoders/decoders with instance methods instead of static ones for several reasons:
- similar to `JSONEncoder`/`JSONDecoder` from `Foundation`
- instance methods make it possible to make en-/decoders "configurable", similarly to JSON coding where you can define a `dateEncodingStrategy` to use when serialising `Date` instances (e.g. `millisecondsSince1970`, `secondsSince1970`, etc.)
- you can inject the corresponding en-/decoder dependency for better testability (to be able to perform method swizzling of static methods you would need Objective-C which would decrease performance a lot)

## Example

To parse a given frame you would first parse the leading `6` bytes as the header to know which frame to parse. Then, depending on the `type`, you can use the corresponding implementation of the `FrameDecoder` protocol:
```swift
var frameBuffer: ByteBuffer = // ...

let headerDecoder = FrameHeaderDecoder()
let frameHeader = try headerDecoder.decode(buffer: &frameBuffer)
switch frameHeader.type {
case .error:
    let frameDecoder = ErrorFrameDecoder()
    let frame = try frameDecoder.decode(header: frameHeader, buffer: &frameBuffer)
    try frame.validate() // optionally validate the frame
// other frame types ...
default: break
}
```

For now I have chosen to use the `throws` syntax, but I'm a bit unhappy about it, since this means any errors thrown are untyped (only constrained to `Error`). What do you think about changing the return type to `Result<ErrorFrame, ErrorFrameError>` instead? This means, the call site knows exactly which type of error that occurs and does not have to have those wildcard catchall error cases. I will leave that up for debate as well.

Another thing I want to discuss is frame validation. For now, the serialisation logic will **not** perform full validation (e.g. `streamId` has to be `0` for certain frames), but I added an extension function for every frame that might perform some validation. What are your thoughts on this?

## Discussion

Here is a quick summary on those three talking points in this PR:
- ~~BinaryKit vs. self made~~
- Typed errors using `Result<Value, Error>` instead of `throws`
- Frame validation


ktoso:
> BinaryKit vs. self made

I would suggest using [Apple / Swift NIO's](https://github.com/apple/swift-nio) [ByteBuffer](https://apple.github.io/swift-nio/docs/current/NIO/Structs/ByteBuffer.html) instead of Data as the core building block for the parsing etc. It is very well suited for this kind of work (en/decoding protocol frames) and is also what we use to power various impls, from http, through grpc ald other protocols in the swift ecosystem. I recommend this also because, realistically, NIO is what is best suited to power the actual networking as well once you get to that part of the implementation -- it abstracts away the underlying networking, so you'll be able to get the best out of running on either apple or server/linux platforms with one impl. You'll also avoid having to roundtrip between Data and other data representations to put the data on the wire this way by sticking to ByteBuffers all the way 👍

OlegDokuka:
@ktoso do you foresee any issues related to interop with other transports that are not using ByteBuffer?

ktoso:
No, it is easier (and potentially allocation-free if possible) to get a Data view data stored in a ByteBuffer - https://github.com/apple/swift-nio/blob/main/Sources/NIOFoundationCompat/ByteBuffer-foundation.swift#L100-L135

It is not possible to avoid allocations going the other way -- if stored as Data and need to write to NIO.

NIO is like Netty in Swift-land which you are well familiar with, and it's what we're building all of the OSS ecosystem tools and libraries on. It automatically offers [Network.framework](https://github.com/apple/swift-nio-transport-services) APIs on Apple platforms as well. One could debate if on iOS one should rather drive using the more high level URLSession, but on server sticking to NIO is _the_ way. And since you'll want to support both stories anyway, sticking to NIO first seems like the better strategy IMHO.

You'll also be in the same boat as the Apple maintained Swift gRPC etc, which is a good boat to be on - sharing the same story about transports.

OlegDokuka:
Gotch!Thank you for confirmation @ktoso! I was thinking the same way but it seemed that Data supports zero-copy as well.

@nkristek do you think we can migrate to ByteBuffer from swift-nio?

nkristek:
I just spoke with @Marcocanc and together we will transition this PR to use `ByteBuffer` instead.

Sources/RSocketCore/Frames/Payload/PayloadFrame.swift
```diff
@@ -0,0 +1,27 @@
...
+ */
+public struct PayloadFrame {
+    /// The header of this frame
+    public let header: FrameHeader
+
+    /// Optional metadata of this frame
+    public let metadata: Data?
+
+    /// Payload for Reactive Streams `onNext`
+    public let payload: Data
```
Samer-Abdulaziz, line 16: why we are using Data for the Payload and metadata and not Bytebuffer?
Actually I saw this repeated in multiple places mainly here and in the interactions frame (FNF...etc)

nkristek: Most of the APIs on iOS work with `Data` (e.g. `JSONDecoder`, `UIImage`, etc.), imho this makes the most sense for the user.

Marcocanc: I think it makes sense to have that as Data. Imo the user of this library should not be required to import nio

OlegDokuka (approved):
LGTM. I guess at this moment it will be hard to judge and I'm pretty sure that some of this will evolve and be changed, but so far soo good.

Sources/RSocketCore/Frames/Setup/SetupFrame.swift
```diff
@@ -102,4 +102,39 @@ public struct SetupFrame {
...
+        if resumeIdentificationToken != nil {
+            flags.insert(.setupResume)
+        }
+        if honorsLease {
+            flags.insert(.setupLease)
+        }
+        self.header = FrameHeader(streamId: streamId, type: .setup, flags: flags)
+        self.majorVersion = majorVersion
+        self.minorVersion = minorVersion
+        self.timeBetweenKeepaliveFrames = timeBetweenKeepaliveFrames
```
Samer-Abdulaziz, line 125: should we add another init without the majorVersion and minorVersion? because it didn't make sense to setup these value in the connectors

Samer-Abdulaziz (approved):
Looks good, we will need to add more init with different params to make easy implementing the connector

Sources/RSocketCore/Frames/Setup/SetupFrame.swift
OlegDokuka: I guess this will be internal API only, so users would not have to deal with that at all

Sources/RSocketCore/Frame Header/FrameHeaderError.swift
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+public enum FrameHeaderError: Swift.Error {
```
OlegDokuka, line 19: just minor, should it be public?

Sources/RSocketCore/FrameError.swift
```diff
@@ -0,0 +1,64 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+public enum FrameError: Swift.Error {
```
OlegDokuka, line 19: should it be public given that we Have RSocket Error as a public api

Sources/RSocketCore/Frames/Error/ErrorFrameError.swift
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+public enum ErrorFrameError: Swift.Error {
```
OlegDokuka, line 19: Should it be internal?

Sources/RSocketCore/Frames/Setup/SetupFrame.swift
Samer-Abdulaziz: yes, what i meant is when we implement the RsocketConnector

Sources/RSocketCore/Payload/PayloadEncoder.swift
```diff
@@ -0,0 +1,38 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+import NIO
+
+public protocol PayloadEncoding {
```
OlegDokuka, line 20: Should internal

Sources/RSocketCore/Payload/PayloadDecoder.swift
```diff
@@ -0,0 +1,49 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+import NIO
+
+public protocol PayloadDecoding {
```
OlegDokuka, line 20: should be internal

PR #9, opened 2021-01-07
provides essential reactive-streams types

OlegDokuka:
Signed-off-by: Niclas Kristek <dev@nkristek.com>

### DIFF
--- a/Sources/RSocketCore/Reactive Streams/SinglePublisher.swift
+++ b/Sources/RSocketCore/Reactive Streams/SinglePublisher.swift
@@ -1,0 +1,19 @@
+/*
+ * Copyright 2015-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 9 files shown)


Sources/RSocketCore/Reactive Streams/SinglePublisher.swift
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+public protocol SinglePublisher: Publisher {
+    
+}
```
nkristek, line 19: Do we need this protocol if we use it just as a sort of namespace to signal the user that this publisher will only ever produce one value (future/promise) or will we add requirements in an upcoming PR?

OlegDokuka: We need that only as a way to define a single value producer in the API. Do you have something in mind to improve that or do it in a different way?

nkristek (approved):
LGTM

Samer-Abdulaziz (approved):

Sources/RSocketCore/Reactive Streams/SinglePublisher.swift
Samer-Abdulaziz: I think this is the similar way we have in JavaScript right ?

ktoso: There's two answers here: 

- EventLoopFuture is "the" future type on the server, however you may not want to expose it since it'll be going away in APIs anyway soon because...
- `async` functions is the actual right answer -- that will land in Swift 6 likely though, so you can't rely on it today yet.


You can keep this for now; there isn't a good standardized future type, but be aware that anything returning this should instead become async functions in the future.

PR #10, opened 2021-01-11
fixes integer overflow while bitwise shifting

AasthaGupta:
This line was doing bitwise shifting by 10 over an 8 bit integer, which will lead to integer overflow. We need to first type cast to UInt16 and then do shifting by 10.


nkristek (approved):
Good catch!

Samer-Abdulaziz (approved):

PR #12, opened 2021-01-12
provides api definition for DuplexConnection and RSocket protocols

OlegDokuka:
This PR introduces the essential protocols for DuplexConnection, RSocket, and some minor thins like Disposable and Closable objects

Signed-off-by: Oleh Dokuka <shadowgun@i.ua>

### DIFF
--- a/Sources/RSocketCore/DuplexConnection.swift
+++ b/Sources/RSocketCore/DuplexConnection.swift
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2015-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+import NIO
+
+/// Represents a connection with input/output that the protocol uses.
+public protocol DuplexConnection: Closeable {
+    ///
+    /// - Returns the assigned `ByteBufAllocator`.
+    ///
+    var allocator: ByteBufferAllocator { get }
+
+    ///
+    /// Return the remote address that this connection is connected to. The returned `SocketAddress`
+    /// varies by transport type
+    ///
+    var remoteAddress: SocketAddress { get }
+
+    ///
+    /// Delivers the given frame to the underlying transport connection. This method is non-blocking
+    /// and can be safely executed from multiple threads. This method does not provide any flow-control
+    /// mechanism.
+    ///
+    /// - Parameters:
+    ///   - streamId: streamId to which the given frame relates
+    ///   - frame: frame with the encoded content
+    func sendFrame(streamId: Int32, frame: ByteBuffer)
+
+    ///
+    /// Send an error frame and after it is successfully sent, close the connection.
+    ///
+    /// - Parameter error: to encode in the error frame
+    func sendErrorAndClose(error: Error)
+
+    ///
+    /// Returns a stream of all frame received on this connection.
+    ///
+    /// Completion:
+    ///
+    /// Returned `Publisher` MUST never emit a completion event
+    ///
+    /// Error:
+    ///
+    /// Returned `Publisher` can error with various transport errors. If the underlying
+    /// physical connection is closed by the peer, then the returned stream from here MUST
+    /// emit an appropriate close error.
+    ///
+    /// Multiple Subscriptions:
+    ///
+    /// Returned `Publisher` is not required to support multiple concurrent subscriptions.
+    /// RSocket will never have multiple subscriptions to this source. Implementations MUST
+    /// emit an appropriate error for subsequent concurrent subscriptions, if they do not
+    /// support multiple concurrent subscriptions.
+    ///
+    /// - Returns Stream of all frames received.
+    ///
+    func receive() -> AnyPublisher<ByteBuffer, Swift.Error>
+}
--- a/Sources/RSocketCore/Reactive Streams/Operators.swift
+++ b/Sources/RSocketCore/Reactive Streams/Operators.swift
@@ -1,0 +1,306 @@
+/*
+ * Copyright 2015-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+internal struct EmptySubscription: Subscription {
+    static let instance = EmptySubscription()
+
+    func request(_ demand: Demand) {
+    }
+
+    func cancel() {
+    }
+}
+
+internal struct ErrorPublisher<Output, Failure: Swift.Error>: SinglePublisher {
+    typealias Output = Output
+    typealias Failure = Failure
+
+    let failure: Failure
+
+    func receive<TSubscriber: Subscriber>(subscriber: TSubscriber) where Failure == TSubscriber.Failure, Output == TSubscriber.Input {
+        subscriber.receive(subscription: EmptySubscription.instance)
+        subscriber.receive(completion: .failure(failure))
+    }
+}
+
+
+internal extension Publisher {
+    /// Wraps this publisher with a type eraser.
+    ///
+    /// Use `eraseToAnyPublisher()` to expose an instance of `AnyPublisher` to
+    /// the downstream subscriber, rather than this publisher’s actual type.
+    /// This form of _type erasure_ preserves abstraction across API boundaries, such as
+    /// different modules.
+    /// When you expose your publishers as the `AnyPublisher` type, you can change
+    /// the underlying implementation over time without affecting existing clients.
+    ///
+    /// The following example shows two types that each have a `publisher` property.
+    /// `TypeWithSubject` exposes this property as its actual type, `PassthroughSubject`,
+    /// while `TypeWithErasedSubject` uses `eraseToAnyPublisher()` to expose it as
+    /// an `AnyPublisher`. As seen in the output, a caller from another module can access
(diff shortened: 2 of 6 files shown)


Sources/RSocketCore/DuplexConnection.swift
```diff
@@ -0,0 +1,75 @@
...
+import NIO
+
+
+/// Represents a connection with input/output that the protocol uses.
+public protocol DuplexConnection: Closeable {
+
+    ///
+    /// - Returns the assigned `ByteBufAllocator`.
+    ///
+    var alloc: ByteBufferAllocator { get }
```
nkristek, line 27: Could we rename this to `allocator`?
```suggestion
    var allocator: ByteBufferAllocator { get }
```

Sources/RSocketCore/DuplexConnection.swift
```diff
@@ -0,0 +1,75 @@
...
+
+    ///
+    /// Delivers the given frame to the underlying transport connection. This method is non-blocking
+    /// and can be safely executed from multiple threads. This method does not provide any flow-control
+    /// mechanism.
+    ///
+    /// - Parameters:
+    ///   - streamId: streamId to which the given frame relates
+    ///   - frame: frame with the encoded content
+    func sendFrame(streamId: Int32, frame: ByteBuffer)
```
nkristek, line 44: Could we use `/** ... */` instead of using multiple `///` for multiline documentation comments?

Sources/RSocketCore/DuplexConnection.swift
```diff
@@ -0,0 +1,75 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+import NIO
+
+
```
nkristek, line 20: Could we have only have one space between imports and definitions?
```suggestion

```

Sources/RSocketCore/Utility/Disposable.swift
```diff
@@ -0,0 +1,39 @@
...
+    ///
+    /// Optionally return `true` when the resource or task is disposed.
+    ///
+    /// Implementations are not required to track disposition and as such may never
+    /// return `true` even when disposed. However, they MUST only return true
+    /// when there's a guarantee the resource or task is disposed.
+    ///
+    /// - Returns `true` when there's a guarantee the resource or task is disposed.
+    ///
+    func isDisposed() -> Bool
```
nkristek, line 38: This could be a variable:
```suggestion
    var isDisposed: Bool { get }
```

Sources/RSocketCore/Utility/Closeable.swift
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+import Foundation
+
+public protocol Closeable: Disposable {
+
+    /// - Returns a `AnySinglePublisher<Never, Swift.Error>` that terminates when the instance is terminated by any reason. Note, in
+    /// case of error termination, the cause of error will be propagated as an error signal through
+    /// `Subscriber#receive(completion: .failure(cause))`. Otherwise, Subscriber#receive(completion: .finished)} will be called.
+    func onClose() -> AnySinglePublisher<Never, Swift.Error>
```
nkristek, line 24: Technically this could be a variable as well, but I'm unsure if it fits the context. It could be implemented as a computed property.
```suggestion
    var onClose: AnySinglePublisher<Never, Swift.Error> { get }
```

Sources/RSocketCore/Reactive Streams/Operators.swift
```diff
@@ -0,0 +1,314 @@
...
+        subscriber.receive(completion: .failure(failure))
+    }
+}
+
+
+internal extension Publisher {
+
+    /// Wraps this publisher with a type eraser.
+    ///
+    /// Use `eraseToAnyPublisher()` to expose an instance of `AnyPublishe`` to
```
nkristek, line 47:
```suggestion
    /// Use `eraseToAnyPublisher()` to expose an instance of `AnyPublisher` to
```

Sources/RSocketCore/Reactive Streams/Operators.swift
```diff
@@ -0,0 +1,314 @@
...
+    ///         print("Successfully cast nonErased.publisher.")
+    ///     }
+    ///     let erased = TypeWithErasedSubject()
+    ///     if let subject = erased.publisher as? PassthroughSubject<Int,Never> {
+    ///         print("Successfully cast erased.publisher.")
+    ///     }
+    ///
+    ///     // Prints "Successfully cast nonErased.publisher."
+    ///
+    /// - Returns: An ``AnyPublisher`` wrapping this publisher.
```
nkristek, line 130:
```suggestion
    /// Wraps this publisher with a type eraser.
    ///
    /// Use `eraseToAnySinglePublisher()` to expose an instance of `AnySinglePublisher` to
    /// the downstream subscriber, rather than this publisher’s actual type.
    /// This form of _type erasure_ preserves abstraction across API boundaries, such as
    /// different modules.
    /// When you expose your publishers as the `AnySinglePublisher` type, you can change
    /// the underlying implementation over time without affecting existing clients.
    ///
    /// The following example shows two types that each have a `publisher` property.
    /// `TypeWithSubject` exposes this property as its actual type, `PassthroughSubject`,
    /// while `TypeWithErasedSubject` uses `eraseToAnySinglePublisher()` to expose it as
    /// an `AnySinglePublisher`. As seen in the output, a caller from another module can access
    /// `TypeWithSubject.publisher` as its native type. This means you can’t change your
    /// publisher to a different type without breaking the caller. By comparison,
    /// `TypeWithErasedSubject.publisher` appears to callers as an `AnySinglePublisher`, so you
    /// can change the underlying publisher type at will.
    ///
    ///     public class TypeWithSubject {
    ///         public let publisher: some SinglePublisher = PassthroughSubject<Int,Never>()
    ///     }
    ///     public class TypeWithErasedSubject {
    ///         public let publisher: some SinglePublisher = PassthroughSubject<Int,Never>()
    ///             .eraseToAnySinglePublisher()
    ///     }
    ///
    ///     // In another module:
    ///     let nonErased = TypeWithSubject()
    ///     if let subject = nonErased.publisher as? PassthroughSubject<Int,Never> {
    ///         print("Successfully cast nonErased.publisher.")
    ///     }
    ///     let erased = TypeWithErasedSubject()
    ///     if let subject = erased.publisher as? PassthroughSubject<Int,Never> {
    ///         print("Successfully cast erased.publisher.")
    ///     }
    ///
    ///     // Prints "Successfully cast nonErased.publisher."
    ///
    /// - Returns: An ``AnySinglePublisher`` wrapping this publisher.
```

Sources/RSocketCore/Reactive Streams/Operators.swift
```diff
@@ -0,0 +1,314 @@
...
+/// A type-erasing publisher.
+///
+/// Use `AnyPublisher` to wrap a publisher whose type has details you don’t want to expose
+/// across API boundaries, such as different modules. Wrapping a `Subject` with
+/// `AnyPublisher` also prevents callers from accessing its `send(_:)` method. When you
+/// use type erasure this way, you can change the underlying publisher implementation over
+/// time without affecting existing clients.
+///
+/// You can use OpenCombine’s `eraseToAnyPublisher()` operator to wrap a publisher with
+/// `AnyPublisher`.
```
nkristek, line 200:
```suggestion
/// Use `AnySinglePublisher` to wrap a publisher whose type has details you don’t want to expose
/// across API boundaries, such as different modules. Wrapping a `Subject` with
/// `AnySinglePublisher` also prevents callers from accessing its `send(_:)` method. When you
/// use type erasure this way, you can change the underlying publisher implementation over
/// time without affecting existing clients.
///
/// You can use `eraseToAnySinglePublisher()` operator to wrap a publisher with
/// `AnySinglePublisher`.
```

Sources/RSocketCore/Reactive Streams/Operators.swift
```diff
@@ -0,0 +1,314 @@
...
```
nkristek, line 146:
```suggestion
/// You can use `eraseToAnyPublisher()` operator to wrap a publisher with `AnyPublisher`.
```

Sources/RSocketCore/Reactive Streams/Operators.swift
```diff
@@ -0,0 +1,314 @@
...
+
+extension AnySinglePublisher: SinglePublisher {
+
+    /// This function is called to attach the specified `Subscriber` to this `Publisher`
+    /// by `subscribe(_:)`
+    ///
+    /// - SeeAlso: `subscribe(_:)`
+    /// - Parameters:
+    ///     - subscriber: The subscriber to attach to this `Publisher`.
+    ///                   once attached it can begin to receive values.
```
nkristek, line 237:
```suggestion
    /// This function is called to attach the specified `Subscriber` to this `SinglePublisher`
    /// by `subscribe(_:)`
    ///
    /// - SeeAlso: `subscribe(_:)`
    /// - Parameters:
    ///     - subscriber: The subscriber to attach to this `SinglePublisher`.
    ///                   once attached it can begin to receive values.
```

Sources/RSocketCore/Utility/Disposable.swift
OlegDokuka: this should rely on an atomic boolean field, thus is not it going to be calculated once if is defined as var?

Sources/RSocketCore/DuplexConnection.swift
OlegDokuka: does that make any difference. I just followed the style defined in swift-nio. e.g -> https://github.com/apple/swift-nio/blob/main/Sources/NIO/Channel.swift#L276

Sources/RSocketCore/Utility/Disposable.swift
Marcocanc: Not if a computed property is used in the implementation
``` swift
var isDisposed: Bool {
    //check if isDisposed and return bool
   return result
}
```

OlegDokuka: gotcha. Thanks for clarification

Sources/RSocketCore/DuplexConnection.swift
nkristek: I think it comes down to personal preference. I'm fine with both, though I have used `/**` previously for multiline comments.

Sources/RSocketCore/Utility/Disposable.swift
```diff
@@ -0,0 +1,39 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+public protocol Disposable {
+
```
nkristek, line 20: Could we also remove the empty lines in the beginning of each protocol to be similar to the SwiftNIO source code?

OlegDokuka: sure

nkristek (approved):
LGTM

PR #13, opened 2021-01-26
WebSocket transport interface 

Samer-Abdulaziz:
This PR includes
1- Folder structure for the transport implementation
2- All the dependencies we will use for the transport
3- Client transport interface
4- Files for transport implementations and public methods to client

We want to get this PR out first so we are aligned on the package structure, dependencies and clientTransport interface. We also want to discuss how the headers support will be implemented 
Currently we have implemented the setHeader method to get headers from the client as

`public func setHeaders(headers: [String: String]) {
        for (key, value) in headers {
            httpHeaders.add(name: key, value: value)
        }
    }`

HTTP headers are not case sensitive. However, setting the key: value as String here will make the headers case sensitive. Should we write a custom implementation to make headers not case sensitive or we can keep it as it is for now?


Sources/RSocketSwiftNIO/wss/WebSocketClientTransport.swift
```diff
@@ -0,0 +1,43 @@
...
+
+import Foundation
+import NIOHTTP1
+import RSocketCore
+
+public final class WebSocketClientTransport: ClientTransport {
+    var httpHeaders = HTTPHeaders()
+    let url: URL
+
+   public init(url: URL) {
```
Marcocanc, line 25: The inset is off here

Sources/RSocketSwiftNIO/wss/WebSocketClientTransport.swift
```diff
@@ -0,0 +1,43 @@
...
+        self.url = url
+    }
+
+    public convenience init(url: String) {
+        self.init(url: URL(string: url)!)
+    }
```
Marcocanc, line 31: Any reason why we should provide this convenience init from string? It's force unwrapping and I'd rather not have this in the framework.

Samer-Abdulaziz: The reason is to give an option to the client to send the endpoint in String format similar to the remaining rsocket clients but I am open to remove it  if it didn't make sense here

Sources/RSocketSwiftNIO/wss/WebSocketClientTransport.swift
Samer-Abdulaziz: fixed, thanks

Sources/RSocketSwiftNIO/wss/WebSocketClientTransport.swift
Marcocanc: However by force unwrapping we are introducing a potential crash here that is not even catchable. I think it's ok to only accept `URL` and delegate the responsibility of providing a valid URL to the imlpementer

Samer-Abdulaziz: Make sense, will remove the String and keep the URL

Samer-Abdulaziz: @Marcocanc  I did removed convenience init

OlegDokuka (approved):
lgtm

Sources/RSocketSwiftNIO/wss/WebSocketClient+Connect.swift
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+import Foundation
+import RSocketCore
+
+public extension WebSocketClientTransport {
+    func connect() -> DuplexConnection {
+        fatalError("connect method is not implemented yet")
+    }
+}
```
nkristek, line 24: Since this file contains the actual implementation of the protocol, could we move the protocol conformance to this file and remove it from the base definition?
```suggestion
public extension WebSocketClientTransport: ClientTransport {
    func connect() -> DuplexConnection {
        fatalError("connect method is not implemented yet")
    }
}
```

Samer-Abdulaziz: Yes, I did add them to one file and this will be part of the next PR

ciychodianda (approved):
lgtm

Marcocanc:
Can this be closed?

dnadoba:
Yes, we already support WebSocket

PR #15, opened 2021-01-28
Add unit tests for frames.

AasthaGupta:
This PR includes unit test for the different frame type. It covers test cases for initialization, encoding, decoding and flag setting.

This will increase the code coverage from 0% to 64%.

Screen Shot 2021-01-28 at 6 16 41 PM


OlegDokuka:
@AasthaGupta that is too early to add tests now. The API will be changed, so adding tests at this moment will just make it more challenging for us to break things when we find them less convenient

Tests/RSocketCoreTests/Frames/PayloadFrameTests.swift
```diff
@@ -0,0 +1,198 @@
...
+        let payloadFrame = PayloadFrame(streamId: 1, fragmentsFollow: true, isCompletion: true, isNext: true, payload: Payload(metadata: payloadMetadata, data: payloadData))
+
+        guard var byteBuffer = try? PayloadFrameEncoder().encode(frame: payloadFrame, using: ByteBufferAllocator()) else {
+            XCTFail("Expected byte buffer to be not nil")
+            return
+        }
+
+        XCTAssertEqual(byteBuffer.readableBytes, 11)
+        XCTAssertEqual(byteBuffer.readBytes(length: byteBuffer.readableBytes), PayloadFrameTests.bytes)
+    }
```
nkristek, line 60: This is just a suggestion and fine if we don't apply it, but we could also define the enclosing method to be throwing:
```suggestion
    func testPayloadEncoder() throws {
        let payloadMetadata = "m".data(using: .utf8)!
        let payloadData = "d".data(using: .utf8)!
        let payloadFrame = PayloadFrame(streamId: 1, fragmentsFollow: true, isCompletion: true, isNext: true, payload: Payload(metadata: payloadMetadata, data: payloadData))

        var byteBuffer = try PayloadFrameEncoder().encode(frame: payloadFrame, using: ByteBufferAllocator())

        XCTAssertEqual(byteBuffer.readableBytes, 11)
        XCTAssertEqual(byteBuffer.readBytes(length: byteBuffer.readableBytes), PayloadFrameTests.bytes)
    }
```

nkristek (commented):
First of all, well done and thank you! I have a few suggestions, otherwise very good!

nkristek:
> @AasthaGupta that is too early to add tests now. The API will be changed, so adding tests at this moment will just make it more challenging for us to break things when we find them less convenient.
> 
> At least, I would deffer mergin that until the internal implementation will be stabilized

Frame logic can be tested imho, it shouldn't change that much anymore. I will add some convenience soon that builds on the existing logic.

OlegDokuka:
@nkristek 
Right, the framing logic itself - true. Encoders - I would disagree. As we discussed, we can add more improvements to the existing codecs API. From my personal experience, writing for unstabilized API end ups in slowing down the whole development and research process

PR #19, opened 2021-02-25
RSocket protocol instead of one function and one closure to create streams

dnadoba:
### Motivation:
`Requester` and `Responder` should use the same protocol to be more in line with implementations of RSocket in other languages.
This means that `Requester` should provide the `RSocket` interface and `Responder` should be given an object which implements the `RSocket` interface.

### Modifications:

- Unification of `StreamOutput` and `StreamInput` into one type called `RStream` (better name suggestions welcome).
- Use `RStream` instead of `AnyPublisher` and related types in `RSocket`
- Use RSocket instead of `(StreamType, Payload, StreamOutput) -> StreamInput` in Responder
- DemultiplexerHandler conforms now to `RSocket`

### Result:

The API is now much simpler and more in line with other languages.


OlegDokuka:
Good job, @dnadoba!

I like the direction it goes!

OlegDokuka (approved):
LGTM

Samer-Abdulaziz (approved):
looks good thanks @dnadoba

Sources/RSocketCore/RSocket.swift
```diff
@@ -17,40 +17,13 @@
...
-    func requestChannel(initialPayload: Payload, payloadsPublisher: AnyPublisher<Payload, Swift.Error>) -> AnyPublisher<Payload, Swift.Error>
-
-    func metadataPush(_ payload: Payload) -> AnySinglePublisher<Never, Swift.Error>
+    func metadataPush(payload: Payload)
+    
+    func fireAndForget(payload: Payload, input: RStream) -> RStream
+    
+    func requestResponse(payload: Payload, input: RStream) -> RStream
+    
+    func stream(payload: Payload, initialRequestN: Int32, input: RStream) -> RStream
```
OlegDokuka, line 26: what is the purpose of having channel behavior (e.g. sending stream) in unidirection requestStream interaction?

Sources/RSocketCore/Streams/Responder.swift
```diff
@@ -17,15 +17,15 @@
 import NIO
 
 internal final class Responder: FrameHandler {
-    private let createStream: (StreamType, Payload, StreamOutput) -> StreamInput
+    private let responderSocket: RSocket
```
nkristek, line 20: This means, the responder/requester have a strong reference to the demultiplexer, but the demultiplexer already has a strong reference to both which in turn creates a memory cycle. The closure was using `[weak] self` in order to not fall into this trap. Since there will be one pipeline per client, this will create a memory leak because clients are not cleaned up after.

nkristek (changes requested):

nkristek (approved):

Sources/RSocketCore/RSocket.swift
dnadoba: An `RStream` can only send data in one direction. A better name would maybe be something like Simplex (opposite of Duplex according to wikipedia)
Only one `RStream` should send data (e.g. onNext) but both sides can still cancel the stream at any time.

As I think about the semantics, the parameter name `input` is misleading because depending from which side you look at it, either from the Requester or Responder side, each `RStream` is an input for the one and an output for the other.
Maybe we should call it `requesterOutput` or `responderInput` and clarify in the documentation that the `RStream` that is given to `stream(payload:initalRequestN:input)` will receive the messages from the responder and the `RStream` returned can be used to send messages to the responder and the only valid message is a cancel frame.

nkristek: It's a good idea to limit the interaction a call site can do on some requests, i.e. requestResponse should only be cancelled, so returning a `Cancellable` should be enough. Together with @dnadoba we thought about changing the return types for the methods to:
- RequestFNF: no return value
- RequestResponse: `Cancellable` (allows for sending cancel)
- RequestStream: `Subscription: Cancellable` (allows for sending demand + cancel)
- RequestChannel: `UnidirectionalStream: Subscription` (is equivalent to the `RStream` protocol)

PR #20, opened 2021-03-04
implements RSocket Protocol

dnadoba:
### Motivation:
`Requester` and `Responder` should use the same protocol to be more in line with implementations of RSocket in other languages.
This means that `Requester` should provide the `RSocket` interface and `Responder` should be given an object which implements the `RSocket` interface.

### Modifications:

- OpenCombine `AnyPublisher`, `SinglePublisher`, `Cancelable`, `Disposable`, etc. is removed
- Use `Cancelable`, `Subscription` and `UnidirectionalStream` instead of `AnyPublisher` and related types in `RSocket`
- Use `RSocket` instead of `(StreamType, Payload, StreamOutput) -> StreamInput` in `Responder`
- Refactor `StreamAdapter` and `StreamFragmenter` into two separate types `RequesterStream` and `ResponderStream` to support new `Cancelable`, `Subscription` and `UnidirectionalStream` instead of `StreamInput` and `StreamOutput`
- add `ThreadSafeStreamAdapter` which converts all stream events (eg. onNext, onCancel, etc,) into frames and forwards them on the right `EventLoop`
- DemultiplexerHandler conforms now to `RSocket`
- implement correct stream termination behaviour 
- move frame fragmentation into `FrameEncoderHandler`

### Result:

The API is now much simpler and more in line with other languages.
streams are terminated i.e. removed from the internal dictionary of active streams
stream events from "clients" are forwarded to the event loop of their corresponding channel


Sources/RSocketCore/Streams/Responder.swift
```diff
@@ -16,57 +16,44 @@
...
-                activeStreams.removeValue(forKey: streamId)
-            }
             return
         }
 
-        let fragmenter = StreamFragmenter(
-            streamId: streamId,
-            maximumFrameSize: maximumFrameSize,
-            createInput: createStream
+        let stream = ResponderStream(
```
nkristek, line 41: Before creating the stream we should check if the initial frame is one of the request frames, otherwise we should drop the frame.

Sources/RSocketCore/Streams/ResponderStream.swift
```diff
@@ -0,0 +1,134 @@
...
+    fileprivate enum StreamKind {
+        case requestResponse(Cancellable)
+        case stream(Subscription)
+        case channel(UnidirectionalStream)
+    }
+    private enum State {
+        
+        case waitingForInitialFragments(RSocket)
+        case active(StreamKind)
+    }
```
nkristek, line 29:
```suggestion
    private enum State {
        case waitingForInitialFragments(RSocket)
        case active(StreamKind)
    }
```

Sources/RSocketCore/Streams/TerminationBehaviour.swift
```diff
@@ -0,0 +1,128 @@
...
+        case .cancel, .complete:
+            guard state != .requesterTerminated else { return true }
+            state = .responderTerminated
+            return false
+        }
+    }
+}
+
+
+
```
nkristek, line 105: Please adjust whitespace to one line between structs

Sources/RSocketCore/Streams/Requester.swift
```diff
@@ -13,20 +13,21 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import NIO
 
-internal final class Requester: FrameHandler {
+internal final class Requester {
     private let sendFrame: (Frame) -> Void
+    private let eventLoop: EventLoop
```
nkristek, line 20: Is it safe to get a strong reference to the `EventLoop`? If yes, should we avoid to schedule work with a strong reference to self? E.g. the following will give the event loop a strong reference to self as well:
```swift
eventLoop.execute { [self] in
     let newId = generateNewStreamId()
     send(frame: RequestFireAndForgetFrameBody(payload: payload).frame(withStreamId: newId))
}
```

dnadoba: It safe in the sense that it will not create a reference cycle.
Retaining `self` in an execute block will create a temporary reference cycle which will be gone once the submitted block has been executed.

In the case of fireAndForget we could create a weak reference to `self` and only send the frame if the requester is still alive.
In all other cases, we need to set the stream id on the `ThreadSafeStreamAdapter` because some user code could call e.g. `onNext` on the `ThreadSafeStreamAdapter` and this would crash if it did not have a stream id.

I made a diagram which visualises the ownership of each object:
Reference Cycle
Strong references are solid lines and weak are dashed lines.

Sources/RSocketCore/Channel Handler/FrameEncoderHandler.swift
```diff
@@ -21,15 +21,24 @@ final class FrameEncoderHandler: ChannelOutboundHandler {
...
-            try frameEncoder.encode(frame: frame, into: &buffer)
-            context.write(wrapOutboundOut(buffer), promise: promise)
+            for fragment in frame.splitIntoFragmentsIfNeeded(maximumFrameSize: maximumFrameSize) {
+                // Todo: performance optimization, we could calculate the actual capacity of the current frame
+                // but for now the buffer will grow automatically
+                var buffer = context.channel.allocator.buffer(capacity: FrameHeader.lengthInBytes)
+                try frameEncoder.encode(frame: fragment, into: &buffer)
+                context.write(wrapOutboundOut(buffer), promise: promise)
+            }
+            context.flush()
```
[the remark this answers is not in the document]
dnadoba: Yes, otherwise the data will not be send to the socket.
This is at least suggested by the documentation of [write](https://apple.github.io/swift-nio/docs/current/NIO/Classes/ChannelHandlerContext.html#/s:3NIO21ChannelHandlerContextC5write_7promiseyAA6NIOAnyV_AA16EventLoopPromiseVyytGSgtF)

Sources/RSocketCore/Error/Error.swift
```diff
@@ -298,3 +316,17 @@ extension Error {
...
+
+extension Error {
+    /// Creates an error frame from `self`. Depending on the error type, it uses the given `streamId` or the connection stream id (stream 0).
+    ///
+    /// This allows the call side to create error frames without knowing whether the error should be sent on the connection or on the specified stream.
+    /// It is especially useful in case the error is later changed from an error that should be sent on the stream instead of on the connection.
+    /// Then the call side would not need to be change, thus can not be forgotten.
+    /// - Parameter streamId: used if it is *not* a connection error
+    /// - Returns: Error Frame
+    internal func asFrame(withStreamId streamId: StreamID) -> Frame {
```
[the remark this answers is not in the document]
dnadoba: Sounds good to me but I chose another name because it actually creates a frame from an error and not from a frame body as the other methods do.
Nevertheless, I agree now that they all should be called the same, I will change them all :)

Sources/RSocketCore/Streams/Responder.swift
dnadoba: hm... the `ResponderStream` does that after receiving a complete frame.
But we could receive a late fragmented payload or some other malicious frame which then would be stored in memory forever.
I will add a check which only creates a stream if it is one of the request frames. :)

Sources/RSocketCore/Channel Handler/FrameEncoderHandler.swift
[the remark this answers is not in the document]
OlegDokuka: That is how it is done in reactor-netty as of now https://github.com/reactor/reactor-netty/blob/master/reactor-netty-core/src/main/java/reactor/netty/channel/MonoSendMany.java#L323

Sources/RSocketCore/Streams/ThreadSafeStreamAdapter.swift
```diff
@@ -0,0 +1,61 @@
...
+        self.eventLoop = eventLoop
+        self.delegate = delegate
+    }
+}
+
+extension ThreadSafeStreamAdapter: UnidirectionalStream {
+    private func send<Body>(_ body: Body) where Body: FrameBodyBoundToStream {
+        eventLoop.execute { [self] in
+            self.delegate?.send(frame: body.asFrame(withStreamId: self.id))
+        }
```
OlegDokuka, line 40: can we do a check if it is already on the eventloop before scheduling something?

```suggestion
        if eventLoop.inEventLoop {
            self.delegate?.send(frame: body.asFrame(withStreamId: self.id))
        } else {
            eventLoop.execute { [self] in
                self.delegate?.send(frame: body.asFrame(withStreamId: self.id))
            }
        } 
```

OlegDokuka (changes requested):
LGTM regardless of a few minor comments

Sources/RSocketCore/Streams/ThreadSafeStreamAdapter.swift
OlegDokuka: I guess the block related to streamId generation has to have the same check as in the TreadSafeStreamAdapter. Having both of them check if it is already on the event loop will guarantee serial message sending and ensure that steamId generation happens before the first message being sent. Otherwise, `eventLoop.execute { [self] in` happens always before any other execution and happens before the adapter is given to the user hands

OlegDokuka (approved):

PR #21, opened 2021-03-09
implements metadata push

nkristek:
This PR adds metadata push handling to both requester (send metadata push) and responder (receive metadata push). The demultiplexer will check the frame type and send all metadata push frames to the responder.

### DIFF
--- a/Sources/RSocketCore/Channel Handler/DemultiplexerHandler.swift
+++ b/Sources/RSocketCore/Channel Handler/DemultiplexerHandler.swift
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import Foundation
 import NIO
 
 /// The role of a connection
@@ -40,7 +41,10 @@ internal struct DemultiplexerRouter {
     
     internal var connectionSide: ConnectionRole
     
-    internal func route(for streamId: StreamID) -> Route {
+    internal func route(for streamId: StreamID, type: FrameType) -> Route {
+        guard type != .metadataPush else {
+            return .responder
+        }
         switch (streamId.generatedBy, connectionSide) {
         case (nil, _):
             return .connection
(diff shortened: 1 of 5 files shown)


dnadoba (approved):

Sources/RSocketCore/Channel Handler/DemultiplexerHandler.swift
```diff
@@ -40,7 +41,10 @@ internal struct DemultiplexerRouter {
...
     internal var connectionSide: ConnectionRole
     
-    internal func route(for streamId: StreamID) -> Route {
+    internal func route(for streamId: StreamID, type: FrameType) -> Route {
+        guard type != .metadataPush else {
+            return .responder
+        }
         switch (streamId.generatedBy, connectionSide) {
         case (nil, _):
             return .connection
```
OlegDokuka, line 50: can we move the type check inside the `.connection` case? 

```suggestion
        switch (streamId.generatedBy, connectionSide) {
        case (nil, _):    
            guard type != .metadataPush else {
                return .responder
            }
            return .connection
```

Sources/RSocketCore/Channel Handler/DemultiplexerHandler.swift
```diff
@@ -80,8 +84,8 @@ internal final class DemultiplexerHandler: ChannelInboundHandler {
 }
 
 extension DemultiplexerHandler: RSocket {
-    func metadataPush(payload: Payload) {
-        fatalError("not implemented")
+    func metadataPush(metadata: Data) {
```
OlegDokuka, line 87: Nice 🚀! It worth doing that. I'm planning to do the same in other impls

OlegDokuka (approved):
LGTM. Added minor suggestion

PR #22, opened 2021-03-10
exposes ChannelPipeline extensions API for RSocket setup

dnadoba:
Add Public API to use `RSocketCore` without `@testable import`.
> Note: This PR should be reviewed & merged after #21
### Motivation:
All `ChannelHandler`s are internal. There is no public API to actually add any `ChannelHandler`s to a `ChannelPipeline`.
This is needed if we want to write performance tests because they need to be build with optimisation and `@testable import` is not available in release configuration.
### Modifications:
Modifications:
- Add extension to `ChannelPipeline` to add RSocket client handlers
- Add extension to `ChannelPipeline` to add RSocket server handlers
- Add extension to `ChannelPipeline` to get requester `RSocket` from a connected connection

### Result:
`Core` can now be used without `@testable import`

### DIFF
--- a/Sources/RSocketCore/Channel Handler/DemultiplexerHandler.swift
+++ b/Sources/RSocketCore/Channel Handler/DemultiplexerHandler.swift
@@ -82,25 +82,3 @@ internal final class DemultiplexerHandler: ChannelInboundHandler {
         }
     }
 }
-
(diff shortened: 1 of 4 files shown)


OlegDokuka (approved):
LGTM

nkristek (approved):
LGTM

PR #25, opened 2021-03-10
implements ConnectionStateHandler

nkristek:
This PR implements a `ConnectionStateHandler` that closes the connection when an error frame is received that has streamId `0`.
It also refactors the route handling of the demultiplexer so that it now supports multiple recipients for a single frame.

### DIFF
--- a/Sources/RSocketCore/Channel Handler/ConnectionStateHandler.swift
+++ b/Sources/RSocketCore/Channel Handler/ConnectionStateHandler.swift
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2015-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


Sources/RSocketCore/Channel Handler/ConnectionStateHandler.swift
```diff
@@ -0,0 +1,58 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import NIO
+
+internal final class ConnectionStateHandler {
```
ciychodianda, line 19: This looks good. Do you plan to add this into the pipeline after the other PR is merged? I don't see where this is added into the pipeline currently.

ciychodianda: Also, why another handler when this could be merged into ConnectionStreamHandler?

ciychodianda: Ah, good point. Makes sense to not do work if there is an error.

OlegDokuka (approved):
LGTM

PR #27, opened 2021-03-11
provides ReactiveSwift compatibility

dnadoba:
ReactiveSwift support
> Note: This PR should be reviewed & merged after #24

### Motivation:
Improve the API for users how want to use ReactiveSwift in their App.

### Modifications:
- Add `RequesterAdapter` & `ResponderAdapter` which uses ReactiveSwift types instead of `Cancellable`, `Subscription` and `UnidirectionStream`
- Add tests 

### Result:
Users can now interact with `SignalProducer`s instead of `Cancellable`, `Subscription` and `UnidirectionStream`

### Note:
requestN is currently not implemented. We first need to decide on how we want to handle it, because ReactiveSwift does not support backpressure/demand.


Sources/RSocketReactiveSwift/Responder.swift
```diff
@@ -0,0 +1,193 @@
...
+}
+
+extension RequestResponseResponder: Cancellable {
+    func onCancel() {
+        disposable.dispose()
+    }
+    func onError(_ error: Error) {
+        // TODO: can we deliver this error to the producer?
+        disposable.dispose()
+    }
```
nkristek, line 100: I think stopping the `SignalProducer` should be enough for now. However, we could expose an optional function on the `RSocket` protocol to receive those errors to log them for example.

Sources/RSocketReactiveSwift/Responder.swift
```diff
@@ -0,0 +1,193 @@
...
+    func onExtension(extendedType: Int32, payload: Payload, canBeIgnored: Bool) {
+        guard !canBeIgnored else { return }
+        let error = Error.invalid(message: "\(Self.self) does not support extension type \(extendedType) and it can not be ignored")
+        output.onError(error)
+        disposable.dispose()
+    }
+    func onRequestN(_ requestN: Int32) {
+        // TODO: Not supported by ReactiveSwift. What should we do with it?
```
nkristek, line 140: As a future improvement (doesn't have to be implemented now), we should make the back pressure behaviour configurable (blocking/queueing/dropping).

nkristek (approved):

Sources/RSocketReactiveSwift/Responder.swift
```diff
@@ -0,0 +1,193 @@
...
+}
+
+fileprivate final class RequestResponseResponder {
+    private let disposable: ReactiveSwift.Disposable
+    private let output: UnidirectionalStream
+    
+    init(producer: SignalProducer<Payload, Swift.Error>, output: UnidirectionalStream) {
+        self.output = output
+        self.disposable = producer.start { (event) in
+            output.send(event: event)
```
OlegDokuka, line 84: Should not we add stronger expectations here? if `SinglaProducer` produces more than one signal, it may break the request-response behavior. Also, request-response has a strong relationship between onNext | onError | onComplete. Due to the spec in this interaction type, it can only be one of those signals, where onNext implicitly means onComplete as well

Tests/RSocketReactiveSwiftTests/RSocketReactiveSwiftTests.swift
```diff
@@ -1,5 +1,297 @@
...
+    func testRequestResponse() {
+        let didReceiveRequest = expectation(description: "did receive request")
+        let didReceiveResponse = expectation(description: "did receive response")
+        
+        let serverResponder = TestRSocket(requestResponse: { payload -> SignalProducer<Payload, Swift.Error> in
+            didReceiveRequest.fulfill()
+            XCTAssertEqual(payload, "Hello World")
+            return SignalProducer { observer, lifetime in
+                observer.send(value: "Hello World!")
+                observer.sendCompleted()
```
OlegDokuka, line 64: do we know how many frames this generates?

OlegDokuka (changes requested):
Nice work! However, there are a few unclear points

Tests/RSocketReactiveSwiftTests/RSocketReactiveSwiftTests.swift
OlegDokuka: can we add a test where the producer produces more than one signal to see if it is canceled right after the first onNext?

dnadoba: Good catch! It tried to send two, I have fixed that.

dnadoba: I have added a new test: `RSocketReactiveSwiftTests.testRequestResponseWithMisbehavingResponderSignalProducerWhichSendsTwoValuesInsteadOfOne`

Sources/RSocketReactiveSwift/Responder.swift
dnadoba: Yes, you are right! I have added an explicit check to only ever send one response.

OlegDokuka (approved):
LGTM!

PR #28, opened 2021-03-12
adds universal bootstrap client API

nkristek:
> Please note: The work on this PR has just started and the client API is not finished.

This PR creates modules that represent the client API.

In order to create a client, the user needs to choose a socket implementation (`Network.framework` or Bercley sockets), a transport protocol implementation (`WebSocket` or `TCP`) and a reactive streams implementation (`Combine` or `ReactiveSwift`).

RSocket

To give an example, if a user wants to use `Network.framework`, `WebSocket` and `ReactiveSwift`, creating an using a client would look like this:
```swift
import Network
import ReactiveSwift
import RSocketReactiveSwift
import RSocketTSChannel
import RSocketWebSocketTransport

    config: .defaultMobileToServer,
    transport: WSTransport(),
    timeout: .seconds(30),
    tlsOptions: NWProtocolTLS.Options()
)

let clientProducer: SignalProducer<ReactiveSwiftClient, Swift.Error> = bootstrap.connect(host: "", port: 80)

let client: Property<ReactiveSwiftClient?> = Property(initial: nil, then: clientProducer.flatMapError { _ in .empty })

let streamProducer: SignalProducer<Payload, Swift.Error> = client.producer.skipNil().flatMap(.latest) { $0.requester.requestStream(payload: .empty) }
```


Tests/RSocketReactiveSwiftTests/RSocketReactiveSwiftTests.swift
```diff
@@ -39,7 +39,7 @@ final class RSocketReactiveSwiftTests: XCTestCase {
             XCTAssertEqual(data, metadata)
         })
         let (_, client) = setup(server: serverResponder)
-        client.requester.rSocket.metadataPush(metadata: metadata)
+        client.requester.reactive.metadataPush(metadata: metadata)
```
OlegDokuka, line 42: I remember, we discussed that this will go away. What is the reason to make things complicated with this extra `reactive` field?

Sources/RSocketReactiveSwift/Requester.swift
```diff
@@ -67,12 +67,13 @@ internal struct RequesterAdapter: RSocket {
 }
 
 extension RSocketCore.RSocket {
-    public var rSocket: RSocket { RequesterAdapter(requester: self) }
+    public var reactive: RSocket { RequesterAdapter(requester: self) }
```
OlegDokuka, line 70: any reasons to keep this?

OlegDokuka (changes requested):
left a few comments to clarify

dnadoba (approved):

OlegDokuka (approved):
LGTM

PR #30, opened 2021-03-18
Keep alive connectionstream

ciychodianda:

### Motivation:

1. Added KeepAlive support

### Modifications:

1. Added support for starting a background task that sends KeepAlive frames at regular intervals
2. Added support to receive KeepAlive frames
3. Terminate KeepAlive background task if it doesn't hear back before maxLifetime is exceeded
4. Haven't added logic to store lastReceivedPosition

### Result:

1. A connection that is opened must stay up with KeepAlive frames being exchanged between peers


Sources/RSocketCore/Channel Handler/ConnectionStreamHandler.swift
```diff
@@ -14,10 +14,53 @@
...
+    typealias InboundIn = Frame
     typealias OutboundOut = FrameBody
-    // TODO
+
+    func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+        let frame = unwrapInboundIn(data)
+        if frame.header.type == .keepalive {
+            lastReceivedTime = Date()
+            if frame.header.flags.contains(.keepAliveRespond) {
+                let responseKeepAliveFrameBody = KeepAliveFrameBody(respondWithKeepalive: false, lastReceivedPosition: 0, data: Data())
```
nkristek, line 40: We need to investigate how we could correctly calculate the `lastReceivedPosition`

Sources/RSocketCore/Channel Handler/ConnectionStreamHandler.swift
```diff
@@ -14,10 +14,53 @@
...
+    }
+
+    func channelActive(context: ChannelHandlerContext) {
+        guard !isKeepAliveStarted, setup.timeBetweenKeepaliveFrames > 0 else { return }
+        lastReceivedTime = Date()
+        keepAliveHandle = context.eventLoop.scheduleRepeatedAsyncTask(initialDelay: .milliseconds(Int64(setup.timeBetweenKeepaliveFrames)), delay: .milliseconds(Int64(setup.timeBetweenKeepaliveFrames))) { task in
+            if Int32(self.lastReceivedTime.timeIntervalSince(Date()) * 1000) > self.setup.maxLifetime {
+                let errorFrame = Frame(header: FrameHeader(streamId: .connection, type: FrameType.error, flags: FrameFlags()), body: FrameBody.error(ErrorFrameBody(error: Error.connectionClose(message: "KeepAlive timeout exceeded"))))
+                context.channel.writeAndFlush(errorFrame, promise: nil)
+                self.keepAliveHandle!.cancel()
```
nkristek, line 54:
```suggestion
                task.cancel()
```

Sources/RSocketCore/Channel Handler/ConnectionStreamHandler.swift
```diff
@@ -14,10 +14,53 @@
...
 
+import Foundation
 import NIO
 
-final class ConnectionStreamHandler: ChannelInboundHandler {
-    typealias InboundIn = FrameBody
+final class ConnectionStreamHandler {
+    private var keepAliveHandle: RepeatedTask?
+    private var isKeepAliveStarted = false
+    private var lastReceivedTime = Date()
```
dnadoba, line 23: `Date` should not be used to schedule timers or measure time because it is bound to the system time which is regularly synced and can be changed by the user manually trough the settings app.
I would recommend to use [`ProcessInfo.processInfo.systemUptime`](https://developer.apple.com/documentation/foundation/processinfo/1414553-systemuptime) which [ultimately calls down to mach_absolute_time on Darwin, and clock_gettime with CLOCK_MONOTONIC on Linux.](https://forums.swift.org/t/recommended-way-to-measure-time-in-swift/33326)

Sources/RSocketCore/Channel Handler/ConnectionStreamHandler.swift
```diff
@@ -14,10 +14,53 @@
...
+            if frame.header.flags.contains(.keepAliveRespond) {
+                let responseKeepAliveFrameBody = KeepAliveFrameBody(respondWithKeepalive: false, lastReceivedPosition: 0, data: Data())
+                let keepAliveFrame = Frame(header: responseKeepAliveFrameBody.header(), body: responseKeepAliveFrameBody.body())
+                context.channel.writeAndFlush(keepAliveFrame, promise: nil)
+            }
+        }
+    }
+
+    func channelActive(context: ChannelHandlerContext) {
+        guard !isKeepAliveStarted, setup.timeBetweenKeepaliveFrames > 0 else { return }
```
dnadoba, line 48: I think this method is called only once from SwiftNIO and `isKeepAliveStarted` is not needed or am I missing something?

Sources/RSocketCore/Channel Handler/ConnectionStreamHandler.swift
```diff
@@ -14,10 +14,53 @@
...
+        lastReceivedTime = Date()
+        keepAliveHandle = context.eventLoop.scheduleRepeatedAsyncTask(initialDelay: .milliseconds(Int64(setup.timeBetweenKeepaliveFrames)), delay: .milliseconds(Int64(setup.timeBetweenKeepaliveFrames))) { task in
```
dnadoba, line 50: The task should be cancelled in `channelInactive(context:)` or, AFAIK, it will run/repeat forever.

Sources/RSocketCore/Channel Handler/ConnectionStreamHandler.swift
```diff
@@ -14,10 +14,53 @@
...
+            if Int32(self.lastReceivedTime.timeIntervalSince(Date()) * 1000) > self.setup.maxLifetime {
+                let errorFrame = Frame(header: FrameHeader(streamId: .connection, type: FrameType.error, flags: FrameFlags()), body: FrameBody.error(ErrorFrameBody(error: Error.connectionClose(message: "KeepAlive timeout exceeded"))))
```
dnadoba, line 52: You can create a `Frame` from an `Error` by just calling `asFrame(withStreamId:)`
```suggestion
                let errorFrame = Error.connectionClose(message: "KeepAlive timeout exceeded").asFrame(withStreamId: .connection)
```

Sources/RSocketCore/Channel Handler/ConnectionStreamHandler.swift
```diff
@@ -14,10 +14,53 @@
...
+                context.channel.writeAndFlush(errorFrame, promise: nil)
+                self.keepAliveHandle!.cancel()
+                self.isKeepAliveStarted = false
+                return context.eventLoop.makeFailedFuture(Error.applicationError(message: "KeepAliveHandler Shutdown"))
+            } else {
+                let keepAliveFrameBody = KeepAliveFrameBody(respondWithKeepalive: true, lastReceivedPosition: 0, data: Data())
+                let keepAliveframe = Frame(header: keepAliveFrameBody.header(), body: keepAliveFrameBody.body())
```
dnadoba, line 59:
```suggestion
                let keepAliveframe = KeepAliveFrameBody(respondWithKeepalive: true, lastReceivedPosition: 0, data: Data()).asFrame()
```

Sources/RSocketCore/ChannelPipeline.swift
```diff
@@ -36,13 +36,14 @@ extension ChannelPipeline {
...
             ),
-            ConnectionStreamHandler(),
+            ConnectionStreamHandler(config: config),
         ])
     }
 }
 
 extension ChannelPipeline {
     public func addRSocketServerHandlers(
+        config: ClientSetupConfig,
```
dnadoba, line 46: ClientSetupConfig is not available at this point in time. We need to wait until the client sends `SetupFrameBody` i.e. `initializeConnection` from `ConnectionEstablishmentHandler` is called.

Sources/RSocketCore/Channel Handler/ConnectionStreamHandler.swift
```diff
@@ -14,10 +14,53 @@
...
 
-final class ConnectionStreamHandler: ChannelInboundHandler {
-    typealias InboundIn = FrameBody
+final class ConnectionStreamHandler {
+    private var keepAliveHandle: RepeatedTask?
+    private var isKeepAliveStarted = false
+    private var lastReceivedTime = Date()
+    private let setup: ClientSetupConfig
+
+    internal init(config: ClientSetupConfig) {
```
dnadoba, line 26: We actually do not need most of `ClientSetupConfig` properties. It would be enough to just initialise `ConnectionStreamHandler` with `timeBetweenKeepaliveFrames` and `maxLifetime`.
In addition, `ClientSetupConfig` is only used on the client. On the server, `SetupInfo` holds the information from the client.

Sources/RSocketCore/ChannelPipeline.swift
```diff
@@ -62,7 +63,7 @@ extension ChannelPipeline {
                         requester: Requester(streamIdGenerator: .server, eventLoop: eventLoop, sendFrame: sendFrame),
                         responder: Responder(responderSocket: responder, eventLoop: eventLoop, sendFrame: sendFrame)
                     ),
-                    ConnectionStreamHandler(),
+                    ConnectionStreamHandler(config: config),
```
dnadoba, line 66: The protocol specification says:
> KEEPALIVE frames MAY be initiated by the server and sent upon application request with the (R)espond flag set.

As I understand it, the server should not send `KeepAliveFrame` periodically, but only on user request. 
Is this correct @OlegDokuka?

Sources/RSocketCore/Channel Handler/ConnectionStreamHandler.swift
```diff
@@ -14,10 +14,53 @@
...
+}
+
+extension ConnectionStreamHandler: ChannelInboundHandler {
+    typealias InboundIn = Frame
     typealias OutboundOut = FrameBody
-    // TODO
+
+    func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+        let frame = unwrapInboundIn(data)
+        if frame.header.type == .keepalive {
```
dnadoba, line 37: An early exit with a guard can improves readability because it makes it clear that the handler only handles keep alive frames and removes one level of indention too.

dnadoba (changes requested):

Sources/RSocketCore/Channel Handler/ConnectionStreamHandler.swift
ciychodianda: Yup, I removed it in the next version that I'll be pushing.

Sources/RSocketCore/ChannelPipeline.swift
ciychodianda: The way it is setup today, both server and client send keepAlive frames. However, I feel server sending keepAlive frames is redundant and can be removed. I kept it this way to keep it similar to other RSocket implementations.

Sources/RSocketCore/Channel Handler/ConnectionStreamHandler.swift
ciychodianda: Good catch.

ciychodianda: Here is how I was thinking about cancel

1. Could happen when timeout is exceeded (handled in code now)
2. When app is backgrounded or removed (on exit), all data structures/tasks related to the app will be cleaned up. 

Wouldn't the keepAlive task be cleaned up on (2)?

dnadoba: [The documentation](https://apple.github.io/swift-nio/docs/current/NIO/Classes/RepeatedTask.html) does not mention anything about automatic cancellation on deallocation.
Looking at the code of `RepeatedTask` and related types, I couldn't find a `deinit` implemantion which does automatic cancellation.

Even if `RepeatedTask` would clean itself up on deallocation, it will never be deallocated because the closure with which `RepatedTask` is initialised, has a strong reference to `self` which creates a reference cycle.

ciychodianda:
Pushed latest changes. keepAlive will only be started from client end now. In ConnectionStreamHandler(), keepAlive task is started in channelActive() only when the SetupFrame is received in channelRead(). This works currently since ChannelActive() is not called up the chain when some of the intermediate handlers are removed. With this change, I feel we need a check like isKeepAlive started to make sure we don't have duplicate tasks started inadvertently (E.g. if context.fireChannelActive() is propagated up the stack)

One of the performance test is failing on my local machine. It has never run on my local and I see there are few changes in that test recently. I'll hold off on modifications until these changes are merged.

Sources/RSocketCore/Channel Handler/ConnectionStreamHandler.swift
```diff
@@ -14,10 +14,81 @@
...
+                context.channel.writeAndFlush(keepAliveFrame, promise: nil)
+                return context.eventLoop.makeSucceededFuture(())
+            }
+        }
+    }
+
+    func channelInactive(context: ChannelHandlerContext) {
+        if keepAliveHandle != nil {
+            keepAliveHandle!.cancel()
+        }
```
dnadoba, line 75: keepAliveHandle?.cancel()

dnadoba (commented):
Some edge cases are not handled properly:
- A client sending multiple setup frames
- A server sending a setup frame to the client

As `ConnectionEstablishmentHandler` already has a mechanism of only doing work once a `SetupFrame` has been received, I would suggest initialising `ConnectionStreamHandler` with all data needed. Everything we need is available in `SetupInfo`.

It will also be good to add some tests. 
Testing things like timers can be hard but SwiftNIO has a good `EventLoop` mock ([`EmbeddedEventLoop`](https://apple.github.io/swift-nio/docs/current/NIO/Classes/EmbeddedEventLoop.html)) which allows us to advance time without actually waiting that the time elapses. It's best used in conjunction with [`EmbeddedChannel`](https://apple.github.io/swift-nio/docs/current/NIO/Classes/EmbeddedChannel.html)
The only thing that we need to mock is our notion of "now". This can be done easily buy injecting a function in the initialiser of the handler:
```swift
private let now: () -> TimeInterval
init(now: @escaping () -> TimeInterval) {
    self.now = now
}
```
If you need any help writing tests, let me now :) 

Reviewing was quite hard because there are still changes in this branch which are not really part of your changes. Please merge master into your branch and try to remove commits that are not from this PR so it will be easier for us to review in the future :)

ciychodianda:
@dnadoba thanks for your help this morning. I made the changes and pushed the latest. I couldn't get the timeout test to work since the channel exits before I can read from it. Could you please take a look at what I might be missing? Thanks.

Sources/RSocketCore/Channel Handler/ConnectionStreamHandler.swift
```diff
@@ -14,10 +14,75 @@
...
-    typealias InboundIn = FrameBody
-    typealias OutboundOut = FrameBody
-    // TODO
+final class ConnectionStreamHandler {
+    private var keepAliveHandle: RepeatedTask?
+    
+    /// receive time in **seconds** of the last keepalive frame
+    private var lastReceivedTime: TimeInterval
+    private var timeBetweenKeepaliveFrames: Int32
+    private var maxLifetime: Int32
```
nkristek, line 26:
```suggestion
    private let timeBetweenKeepaliveFrames: Int32
    private let maxLifetime: Int32
```

Sources/RSocketCore/Channel Handler/ConnectionStreamHandler.swift
```diff
@@ -14,10 +14,75 @@
...
+    typealias InboundIn = Frame
+    typealias OutboundOut = Frame
+
+    func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+        let frame = unwrapInboundIn(data)
+        switch frame.body {
+        case let .keepalive(body):
+            lastReceivedTime = now()
+            if body.respondWithKeepalive {
+                let keepAliveFrame = KeepAliveFrameBody(respondWithKeepalive: false, lastReceivedPosition: 0, data: Data()).asFrame()
```
nkristek, line 54: Let's document it with a comment that `lastReceivedPosition` is set to `0` statically because we don't support resumability yet:
```suggestion
                // we don't support resumability yet so we send always send lastReceivedPosition as 0
                let keepAliveFrame = KeepAliveFrameBody(respondWithKeepalive: false, lastReceivedPosition: 0, data: Data()).asFrame()
```

ciychodianda (commented):
looks good

dnadoba (approved):

nkristek (approved):

PR #31, opened 2021-03-22
adds tests that use Network.Framework

dnadoba:
Add tests which uses Network.framework instead of Berkeley Sockets as the underlying channel implementation.

### Motivation:
Measure Performance of Network.Framework and see if it generally works.

### Modifications:
- Add class NetworkFrameEndToEndTests which subclasses EndToEndTests
- overwrite `setupClient` and `setupSever` to bootstrap Network.Framework 
- performance tests now limit parallelism (e.g. concurrent requests in flight) to better model real world usage

### Result:
End-to-End & Performance tests do now use Network.Framework as well

#### Performance Tests Results:
- **macOS**: Network.Framework is **~3x slower** than Berkeley Sockets
- **iOS**: Swift Packages do not support running tests on real iOS device, only on simulator where performance tests are not meaningful
- **Linux**: Network.Framework is not available on Linux

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -27,6 +27,15 @@
           "revision": "de1c80ad1fdff1ba772bcef6b392c3ef735f39a6",
           "version": "1.8.0"
         }
+      },
+      {
+        "package": "swift-nio-transport-services",
+        "repositoryURL": "https://github.com/apple/swift-nio-transport-services",
+        "state": {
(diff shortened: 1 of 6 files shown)


nkristek (approved):

OlegDokuka (approved):
LGTM

PR #32, opened 2021-03-23
improves and adds more tests

dnadoba:
improve existing tests and tests onError
### Motivation:
improving test suite and code coverage
### Modifications:
- Fail if we receive an unexpected event
- Fail if we receive an unexpected late frame
- Test `onError`
- Refactor server and client bootstrap in EndToEndTests by moving it from the test case into one method

### Result:
slightly higher test coverage and cleaner test cases


Sources/RSocketCore/Streams/Requester.swift
```diff
@@ -22,15 +22,18 @@ internal final class Requester {
     private let eventLoop: EventLoop
     private var streamIdGenerator: StreamIDGenerator
     private var activeStreams: [StreamID: RequesterStream] = [:]
+    private var lateFrameHandler: ((Frame) -> ())?
```
nkristek, line 25:
```suggestion
    private let lateFrameHandler: ((Frame) -> ())?
```

Sources/RSocketCore/Streams/Responder.swift
```diff
@@ -21,15 +21,17 @@ internal final class Responder {
     private let sendFrame: (Frame) -> Void
     private var activeStreams: [StreamID: ResponderStream] = [:]
     private let eventLoop: EventLoop
-    
+    private var lateFrameHandler: ((Frame) -> ())?
```
nkristek, line 24:
```suggestion
    private let lateFrameHandler: ((Frame) -> ())?
    
```

nkristek (approved):

PR #33, opened 2021-03-25
adds fragmentation tests

dnadoba:
Test frame fragmentation and assembly. 

### Motivation:

Improve test coverage

### Modifications:
- make it convenient to initialise payload from a string for testing purposes
- write a bunch of test cases

### Result:
Frame fragmentation and assembly is now tested.
Note: An End-to-End test which does fragmentation is still missing because this would result in merge conflicts with #31.


OlegDokuka (approved):
LGTM

PR #36, opened 2021-03-25
optimizes reading and writing 24 bit integers

nkristek:
This PR optimizes the reading and writing of 24 bit integers needed for metadata length en-/decoding.

### Motivation:

Reading/writing separate `UInt8` & `UInt16` is more efficient than having to allocate an additional array of `[UInt8]`.

### Modifications:

I've added new extension methods on `ByteBuffer` which follow the naming scheme typically for SwiftNIO, namely:
- `setUInt24(_:at:endianness:as:)`
- `writeUInt24(_:endianness:as:)`
- `getUInt24(at:endianness:)`
- `readUInt24(endianness:)`

### Result:

The new en-/decoding methods are currently only used in the `PayloadEncoder` and `PayloadDecoder`.


OlegDokuka (approved):
LGTM

Tests/RSocketCoreTests/ByteBufferUInt24Tests.swift
```diff
@@ -0,0 +1,129 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import NIO
+@testable import RSocketCore
+import XCTest
+
+final class ByteBufferUInt24Tests: XCTestCase {
+    func testSetUInt24WithUInt8BigEndian() throws {
```
dnadoba, line 22: AFAICT, all test cases do not throw. Any reason to define them as throwing?

Sources/RSocketCore/Utility/ByteBuffer+UInt24.swift
```diff
@@ -0,0 +1,91 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import NIO
+
+extension ByteBuffer {
+    @discardableResult
+    @inlinable
+    internal mutating func setUInt24<T: FixedWidthInteger>(
```
dnadoba, line 22: We may want to constrain it to `UnsignedInteger` as well.
What would it mean to write an `SignedInteger` eg. `Int32` as an `UInt24`?

Sources/RSocketCore/Utility/ByteBuffer+UInt24.swift
```diff
@@ -0,0 +1,91 @@
...
+        _ integer: T,
+        at index: Int,
+        endianness: Endianness = .big,
+        as: T.Type = T.self
+    ) -> Int {
+        let mostSignificant: UInt16
```
dnadoba, line 28: We may want to add a precondition, or at least an assert, to check if the given integer fits in a `UInt24` without losing data.

dnadoba (approved):
Well done! Did you check the assembly code of these methods? I tried to but godbolt.org does not support swift packages yet and ByteBuffer is part of SwiftNIO, so I wasn't able to do a quick check. Apart from a few minor things, LGTM.

Sources/RSocketCore/Utility/ByteBuffer+UInt24.swift
nkristek: Good idea!

Sources/RSocketCore/Utility/ByteBuffer+UInt24.swift
nkristek: Good catch

PR #37, opened 2021-03-26
improves fragmentation impl and adds e2e tests for it

dnadoba:
Test Fragmentation and fix frame assembly bug in `Responder` for stream & channel payload frames

### Motivation:

Increase Test coverage and fix bugs

### Modifications:

- change default maximum fragmentation size to 2^24 (I think this is the default in Java too)
- assert that encoding frames do not throw 
- responder now forwards assembled frame instead of only the last fragment (bug fix)
-  add a bunch of end to end tests with frames that need to be split up into fragments and needs to be reassembled on the other end 

### Result:
- fragment reassembly now works correctly for channel & stream payload frames
- higher test coverage


Tests/RSocketCoreTests/EndToEndTests.swift
```diff
@@ -68,16 +69,28 @@ class EndToEndTests: XCTestCase {
...
+                // uncomment if you want to see all frames send and received
+//                .flatMap {
+//                    channel.pipeline.handler(type: FrameDecoderHandler.self).flatMap {
+//                        channel.pipeline.addHandler(DebugInboundEventsHandler(), name: "server receiving", position: .after($0))
+//                    }.flatMap {
+//                        channel.pipeline.handler(type: FrameEncoderHandler.self).flatMap {
+//                            channel.pipeline.addHandler(DebugOutboundEventsHandler(), name: "server sending", position: .after($0))
+//                        }
+//                    }
+//                }
```
nkristek, line 86: Do we want to add an extension method on pipeline for this? I can imagine this might be useful for other places as well.

Tests/RSocketCoreTests/NetworkFrameworkEndToEndTests.swift
```diff
@@ -72,7 +77,9 @@ final class NetworkFrameworkEndToEndTests: EndToEndTests {
                     channel.pipeline.addRSocketClientHandlers(
                         config: config,
                         responder: responderSocket,
-                        connectedPromise: connectedPromise
+                        maximumFrameSize: maximumFrameSize, connectedPromise: connectedPromise,
```
nkristek, line 80: Could you please add a line break before `connectedPromise`?

nkristek (approved):

OlegDokuka (approved):
LGTM

PR #38, opened 2021-03-26
fixes performance tests on CI

dnadoba:
performance tests now actually run on CI
### Motivation:
Performance tests were skipped because they were build in debug mode but with optimisations turned on
### Modifications:
use `--configuration=release` instead of `-Xswiftc="-O"`

### Result:
Tests are build in release configuration and are executed


OlegDokuka (approved):
LGTM

PR #42, opened 2021-03-31
Fix handling of isNext flag when fragmenting/reassembling frames

nkristek:
This PR changes the behaviour so the `isNext` flag is no longer used to determine if a frame starts a new set of fragments.

### Motivation:

Other implementations of the RSocket protocol expect all fragments to have the `isNext` flag instead of just the first one.

### Modifications:

This PR modifies both the fragmentation, so all frames get the same `isNext` value, and the reassembly, so `isNext` is used to determine if a frame should be fragmented.

### Result:

The fragmentation/reassembly logic is compatible with other RSocket implementations.


Tests/RSocketCoreTests/PayloadFragmentationTests.swift
```diff
@@ -321,36 +321,4 @@ class PayloadFragmentationTests: XCTestCase {
         XCTAssertEqual(assembler.process(frame: fragments[safe: 0]), .incomplete)
         XCTAssertTrue(assembler.process(frame: Error.canceled(message: "cancel").asFrame(withStreamId: .connection))?.isError)
     }
-    
-    func testReceiveNewRequestBeforeReceivingAllFragmentsShouldResultInAnError() {
```
dnadoba, line 325: I think we should still have a test but just not for PAYLOAD. A Frame type other than PAYLOAD e.g. REQUEST_RESPONSE is still a sign of a completely new set of fragments. What do you think?

Sources/RSocketCore/Streams/Defragmentation/FragmentedFrameAssembler.swift
```diff
@@ -55,6 +49,15 @@ internal struct FragmentedFrameAssembler {
...
                     return .error(reason: reason)
                 }
+
+            case (false, true):
+                guard fragments == nil else {
+                    return .error(reason: "Got (C)omplete only frame but there is a current set of fragments to extend")
+                }
+                return .complete(frame)
+
+            case (false, false):
```
OlegDokuka, line 59: 👍

OlegDokuka (approved):
LGTM

dnadoba (approved):

PR #43, opened 2021-04-02
WebSocket Transport

dnadoba:
add basic WebSocket transport support

### Motivation:
make Twitter and Timer examples work

### Modifications:
- generate random `Sec-WebSocket-Key` as defined in [RFC6455](https://tools.ietf.org/html/rfc6455#section-4.1) 
- add WebSocket KeepAlive support by responding to WebSocket ping frames from server with pong frame
- forward inbound binary WebSocket frame as `ByteBuffer`
- forward outbound `ByteBuffer` as binary WebSocket frame
- generate random mask key for outbound WebSocket frames
- add support for user defined HTTP headers
- add support for user defined URI
- simplify all examples by using blocking calls

### Result:
Twitter and Timer examples do work now

##### WebSocket text frame
The current implementation just ignores text frames. How do other implementations handle text frames?

##### WebSocket Fragmentation
The current implementation does not support WebSocket fragmentation. I don't know if we need this because RSocket already supports fragmentation but I could imagine that an intermediate WebSocket Proxy might fragment frames by accident so it might be worth to add it just to be sure.

##### WebSocket close frame
The current implementation does not handle WebSocket close frames. We definitely need to tackle this in the future.


Sources/RSocketCore/Channel Handler/SetupWriter.swift
```diff
@@ -27,8 +27,15 @@ internal final class SetupWriter: ChannelInboundHandler, RemovableChannelHandler
...
-    
+    func handlerAdded(context: ChannelHandlerContext) {
+        if context.channel.isActive {
+            onActive(context: context)
+        }
+    }
     func channelActive(context: ChannelHandlerContext) {
+        onActive(context: context)
+    }
+    private func onActive(context: ChannelHandlerContext) {
```
nkristek, line 38: Can we add one empty line between functions please?

Sources/RSocketCore/Client/ClientBootstrap.swift
```diff
@@ -19,5 +19,10 @@ import NIO
 public protocol ClientBootstrap {
     associatedtype Client
     associatedtype Responder
-    func connect(host: String, port: Int, responder: Responder?) -> EventLoopFuture<Client>
+    func connect(
+        host: String,
+        port: Int,
+        uri: String,
```
nkristek, line 25: Since this parameter is not needed for the TCP transport, we could define the connection configuration as an associated type, so websocket would require different config than TCP. WDYT?

Sources/RSocketCore/Channel Handler/SetupWriter.swift
dnadoba: Sure, should we add something like [swift-lint](https://github.com/realm/SwiftLint) to this project?

Sources/RSocketCore/Client/ClientBootstrap.swift
dnadoba: As noted in the PR description, there are multiple options. I actually like the associated type solution the most.

nkristek (approved):
LGTM!

ciychodianda (approved):
Looks good!

PR #45, opened 2021-04-09
reworks client configuration

dnadoba:
Rethink client configuration

### Motivation:
client configuration was not easy and not documented. The previous  syntax made it not possible to change a single property without create a temporary variable.
### Modifications:
- rename `ClientSetupConfig` to `ClientConfiguration`
- remove setup payload from configuration and move it to the `connect` method on `ClientBoostrap`
- group related configurations into their own type
  - e.g. `timeBetweenKeepaliveFrames` and `maxLifetime` are now in one type called `Timeout` which makes the strong relationship clear and allows for predefined named defaults which only include the relevant properties (e.g. `.disabled`, `.mobileToServer` and `.serverToServer`)
- make incoming and outgoing fragment size limits part of client configuration
- add a nice syntax for changing part of the configuration

Swift 5.3
```swift
    transport: WSTransport(),
    config: ClientConfiguration.mobileToServer
        .set(\.encoding.metadata, to: .rsocketRoutingV0)
        .set(\.encoding.data, to: .json)
)
```
<details>
  <summary>Swift 5.4</summary>

[Swift 5.4 supports implicit member chains](https://github.com/apple/swift-evolution/blob/main/proposals/0287-implicit-member-chains.md)

```swift
    transport: WSTransport(),
    config: .mobileToServer
        .set(\.encoding.metadata, to: .rsocketRoutingV0)
        .set(\.encoding.data, to: .json)
)
```
</details>

### Result:
If we want to customise the configuration, we do no longer need to specify all values.
We can now customise each property or group of related properties individually, however we like.
Instead of providing all values: 
```swift
    config: ClientSetupConfig(
        timeBetweenKeepaliveFrames: 30_000,
        maxLifetime: 60_000,
        metadataEncodingMimeType: "message/x.rsocket.routing.v0",
        dataEncodingMimeType: "application/json"
    ),
    transport: WSTransport()
)
```
we can now just overwrite the values we actually want to change:
```swift
    transport: WSTransport(),
    config: .mobileToServer
        .set(\.encoding.metadata, to: .rsocketRoutingV0)
        .set(\.encoding.data, to: .json)
)
```
In addition, all configuration properties are now documented and fragment size limits are configurable.


Sources/RSocketCore/Channel Handler/FrameEncoderHandler.swift
```diff
@@ -22,9 +22,9 @@ final class FrameEncoderHandler: ChannelOutboundHandler {
 
     private let frameEncoder = FrameEncoder()
     
-    private let maximumFrameSize: Int32
+    private let maximumFrameSize: Int
```
nkristek, line 25: Since frame size can only be as big as `Int32`, why did we change it to `Int`?

Sources/RSocketCore/Client/ClientConfiguration.swift
```diff
@@ -0,0 +1,172 @@
...
+        limits: Limits = .default
+    ) {
+        self.timeout = timeout
+        self.encoding = encoding
+        self.limits = limits
+    }
+}
+
+extension ClientConfiguration {
+
```
nkristek, line 130: Let's remove this empty line:
```suggestion
```

Sources/RSocketCore/Utility/MIMEType.swift
```diff
@@ -0,0 +1,53 @@
...
+/// ```
+public struct MIMEType: RawRepresentable, Hashable {
+    public private(set) var rawValue: String
+    public init(rawValue: String) {
+        self.rawValue = rawValue
+    }
+}
+
+extension MIMEType {
+
```
nkristek, line 41:
```suggestion
```

Sources/RSocketCore/Channel Handler/FrameEncoderHandler.swift
dnadoba: We need to convert it to an Int anyway to do anything useful with it. Int32 also does not capture the full range restriction and still allows invalid values e.g. -1 or Int32.max.
[Int is the recommended general type for integers](https://docs.swift.org/swift-book/LanguageGuide/TheBasics.html):
>Use UInt only when you specifically need an unsigned integer type with the same size as the platform’s native word size. If this isn’t the case, Int is preferred, even when the values to be stored are known to be nonnegative. A consistent use of Int for integer values aids code interoperability, avoids the need to convert between different number types, and matches integer type inference, as described in Type Safety and Type Inference.

Another good real world example is ByteBuffer. It uses an Int as an Index, even if it can't store more than 2^32 bytes.
Another one is ByteToMessageHandler, which uses and Int as a max size, even if it is limited by the size restriction of ByteBuffer.

If we really want to limit `maximumFrameSize`, we need validate them manually because no type in swift can express our valid value range.

nkristek: Ok fair point.

nkristek (approved):

OlegDokuka (approved):
LGTM

PR #46, opened 2021-04-13
improves frame debug description

dnadoba:
Remove duplicate storage of frame flags

### Motivation:
improve the debug description of `Frame`

### Modifications:
- remove duplicate storage of flags from `Frame`
- delay header generation to decoding phase
- move fragmentFollows to FrameBody

### Result:
All frame flags are now all consistently stored in FrameBody. 
Flags are now no longer redundantly store in the header and in the body.

This already improves the debug description of a Frame from:
```
Frame(header: RSocketCore.FrameHeader(streamId: .connection, type: RSocketCore.FrameType.cancel, flags: RSocketCore.FrameFlags(rawValue: 0)), body: RSocketCore.FrameBody.cancel(RSocketCore.CancelFrameBody()))
Frame(header: RSocketCore.FrameHeader(streamId: 7, type: RSocketCore.FrameType.payload, flags: RSocketCore.FrameFlags(rawValue: 288)), body: RSocketCore.FrameBody.payload(RSocketCore.PayloadFrameBody(isCompletion: false, isNext: true, payload: RSocketCore.Payload(metadata: Optional(4 bytes), data: 11 bytes))))
```
to
```
Frame(streamId: .connection, body: RSocketCore.FrameBody.cancel(RSocketCore.CancelFrameBody()))
Frame(streamId: 7, body: RSocketCore.FrameBody.payload(RSocketCore.PayloadFrameBody(fragmentFollows: false, isCompletion: false, isNext: true, payload: RSocketCore.Payload(metadata: Optional(4 bytes), data: 11 bytes))))
```

Flags are now no longer printed twice and can be interpreted without manual converting into binary and looking into the specification.

This is only the beginning and I will continue improving the debug description.
This PR is already big enough and can be reviewed independently.


Sources/RSocketCore/Frames/FrameBody/Request Stream/RequestStreamFrameBody.swift
```diff
@@ -15,7 +15,10 @@
  */
 
 /// Request a completable stream
-internal struct RequestStreamFrameBody: Hashable {
+internal struct RequestStreamFrameBody: Hashable, FragmentableFrameBody {
+    /// If true, this is a fragment and at least another payload frame will follow
+    internal var fragmentFollows: Bool = false
```
nkristek, line 20: Do we want to declare all properties as `var` or all as `let`? In order for the `set` method to work I would go for the former, but I would not mix both options.

dnadoba: I think it is a good idea to make them all `var`s because all frame related types are value types. The use site can then decide if they want a `let` or a `var`.

nkristek (approved):

OlegDokuka (approved):
LGTM

PR #48, opened 2021-04-15
makes Error extensible

dnadoba:
Allows user code to add custom error codes 

### Motivation:
A user should be able to easily define and use custom application errors

### Modifications:
- `Error` is now a struct instead of an enum with two stored properties `code` and `message`
- `Error.Code` is a new struct which is a thin wrapper around a UInt32 which allows to add custom error codes through extensions

### Result:
A user can now define custom errors by writing two extensions:
```
extension Error.Code {
    static let myCustomError = Self(rawValue: 0x00000301)
}
extension Error {
    static func myCustomError(message: String) -> Self { .init(code: .myCustomError, message: message ) }
}
```
Custom Errors can be used in the same places as predefined errors and behave the same too.

### DIFF
--- a/Sources/RSocketCore/Error/Error.swift
+++ b/Sources/RSocketCore/Error/Error.swift
@@ -14,238 +14,28 @@
  * limitations under the License.
  */
 
-/**
- Errors are used on individual requests/streams as well
- as connection errors and in response to `SETUP` frames.
- */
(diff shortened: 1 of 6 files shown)


nkristek (approved):

OlegDokuka (approved):
LGTM

PR #50, opened 2021-05-05
adds NIOWebSocketFrameAggregator to buffer fragmented WS frames

dnadoba:
### Motivation:

We want support frame fragmentation at the WebSocket level

### Modifications:

add NIOWebSocketFrameAggregator to channel pipeline in front of our own channel handlers

### Result:

Fragmented WebSocket Frames are now supported.

### Discussion

What are good default values for `minNonFinalFragmentSize` and `maxAccumulatedFrameCount` for WebSocket?
The equivalent of `minNonFinalFragmentSize` in RSocket would be 64 bytes but I don't know if this is appropriate for WebSocket as well.
I have set `minNonFinalFragmentSize` by default to `0` and `maxAccumulatedFrameCount` to `Int.max` for now.

`minNonFinalFragmentSize` and `maxAccumulatedFrameCount` are part of `WSTransport` and not `ClientConfiguration` because they are specific to WebSocket. I kind of want them to be part of `ClientConfiguration`. I initially thought that we could just add these properties to `ClientConfiguration.Fragmentation` and even use the same values for WebSocket fragmentation and RSocket fragmentation. But that would mean one can not configure fragmentation separately for WebSocket and for RSocket... Maybe that's good thing, less configuration one has to worry about is in general a good thing. However, I can't say for sure that RSocket fragmentation and WebSocket fragmentation should always have the same buffer limits.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -24,8 +24,8 @@
         "repositoryURL": "https://github.com/apple/swift-nio",
         "state": {
           "branch": null,
-          "revision": "3be4e0980075de10a4bc8dee07491d49175cfd7a",
-          "version": "2.27.0"
+          "revision": "6befe13e234e7fa4be6e0e9d8f9d06ee18bb589c",
+          "version": null
         }
       },
       {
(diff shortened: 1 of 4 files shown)


nkristek (approved):

OlegDokuka (approved):
LGTM

PR #53, opened 2021-06-25
adds Encode and Decode Pipelines

dnadoba:
This PR introduces a novel way to define encode and decode pipelines in a type-safe and composable way.

### Let's start with some Examples

This is a simple request definition for a Fire and Forget Request which directly encodes routing information in its metadata and encodes its data as JSON:

```swift
FireAndForget<Metrics> {
    Encoder()
        .encodeStaticMetadata(["metrics"], using: .routing)
        .encodeData(using: JSONDataEncoder(type: Metrics.self))
}
```
Fire and Forget requests do only send payloads and never receive any. They therefore only need to define an encoder.
All other Request types can also receive payloads and need to define a decoder as well.

This is a description of a Request Stream request which uses Composite Metadata to encodes a route. Data is encoded and decoded as JSON:

```swift
RequestStream<ISIN, Price> {
    Encoder()
        .useCompositeMetadata()
        .encodeStaticMetadata(["price"], using: .routing)
        .encodeData(using: JSONDataEncoder(type: ISIN.self))
    Decoder()
        .useCompositeMetadata()
        .decodeData(using: JSONDataDecoder(type: Price.self))
}
```

With the use of the `Coder` convenience type we can simplify it a bit:

```swift
RequestStream<ISIN, Price> {
    Coder()
        .useCompositeMetadata()
        .encodeStaticMetadata(["price"], using: .routing)
        .encodeData(using: JSONDataEncoder(type: ISIN.self))
        .decodeData(using: JSONDataDecoder(type: Price.self))
}
```

This API also supports more advanced use-cases where the caller wants to encode/decode Metadata:

```swift
RequestResponse<(([String], MIMEType), (MIMEType, Price)), (([String], [MIMEType], MIMEType), Stock)> {
    Coder()
        .useCompositeMetadata()
        .encodeData {
            JSONDataEncoder(type: Price.self)
        }
        .encodeMetadata {
            RoutingEncoder()
            DataMIMETypeEncoder()
        }
        .decodeData(using: JSONDataDecoder(type: Stock.self))
        .decodeMetadata {
            RoutingDecoder()
            AcceptableDataMIMETypeDecoder()
            DataMIMETypeDecoder()
        }
}
```
In this advanced case, the metadata is is added as a tuple to the input and output types which makes it available to the caller.

For more examples please look at the `RequestsExamples.swift` under `Sources/RSocketCore/Extensions`.

### Architecture
The Architecture is inspired by the Lazy Sequence/Collection types from the Standard Library, Combine and SwiftUI. It uses nested generics to encode the complete pipeline in the type system. This enables us to chain operators. In addition, the compiler can optimize each pipeline by generating a specialized version because the pipeline is encoded as a type and is completely visible to the compiler.

At the core are two protocols: `EncoderProtocol` and `DecoderProtocol`. 
```swift
public protocol EncoderProtocol {
    associatedtype Metadata
    associatedtype Data
    mutating func encode(
        metadata: Metadata,
        data: Data,
        mimeType: ConnectionMIMEType
    ) throws -> Payload
}
public protocol DecoderProtocol {
    associatedtype Metadata
    associatedtype Data
    mutating func decode(
        _ payload: Payload,
        mimeType: ConnectionMIMEType
    ) throws -> (Metadata, Data)
}
```
These protocols are implemented by operators like `Decoders.DataDecoder` or `Decoders.Map` which usually wrap another decoder. All of those operators provide an extension method on `EncoderProtocol` and `DecoderProtocol`.

To construct a new decode or encode pipeline, we initialize a `Decoder()` or `Encoder()`. Their associated Metadata and Data types are constrained to match those of the `Payload` type (`Foundation.Data?` for Metadata and `Foundation.Data` for Data). We then chain different operators to transform the Metadata and Data to a type we want to consume. At the end we can erase the actual type by calling `.eraseToAnyDecoder()` or `.eraseToAnyEncoder()`.

All conforming types are value types and can store and mutate state. `encode(metadata:data:mimeType)` and `decode(_:mimetype)` are therefore mutating methods. This is, for example, used by the per Stream Data MIME Type Extension. The specification says that the data MIME Type is only send once for the initial payload. Subsequent payloads do not contain the Data MIME Type Metadata and we therefore need to remember the initial MIME Type to decode the data with the right decoder.

All Extension Metadata Encoders and Decoders are not actually implemented. However, the types do already exists to prove that the Decoder/Encoder API would actually be usable with all RSocket Extensions. Metadata and Data Encoders/Decoders follow a similar architecture as the the `Decoder/EncoderProtocol`. Users can create their own implementation and use it at the same places as the builtin encoders and decoders.

Some more advanced and convenience features that are planned but not yet implemented:
- Optionally decode/encode Metadata
- improved syntax for dynamic encoding/decoding of metadata
- custom route type instead of [String] which conforms to ExpressibleByStringLiteral and ExpressibleByArrayLiteral
- Currently we only support up to three different Metadata/Data Encoders/Decoder. Swift does not yet support Variadic Generics and we need to manually implement a different type for each number of decoders/encoders much like SwiftUI does for their Group Views. Not too much work but just a lot of lines of code.

### ErasableMetadata and Result Builders
`ErasableMetadata` is a marker protocol used by `CoderBuilder`, `EncoderBuilder` or `DecoderBuilder` to automatically erase or preserve Metadata. If a type conforms to `ErasableMetadata` like `Data?` or `[CompositeMetadata]` and is the last Metadata, it is erased i.e. ignored by decoders and initialized with a default value by encoders. If Metadata is another type, it is automatically preserved and available by the caller i.e. when sending a request you also need to provide the metadata and the response contains the Metadata in addition to the Data. A user can always manually opt-out or opt-in by using `preserveMetadata()` or `eraseMetadata()`/`setMetadata(_:)`.

### Naming
Naming is difficult. I would like to rename `DecoderProtocol` and `EncoderProtocol` to `PayloadDecoder` and `PayloadEncoder` but we already have existing types with that name and we first need to rename those.
Different naming suggesting are as always welcome :) 

### Next Steps
As this PR is already huge I did removed some code to keep it focused. 
In a follow-up PR I will make it possible to actually make Requests with this new API. 
In addition, this API is designed to not only be used for making request but also to accept incoming request. In another follow-up PR I will make it possible to define responder handlers using this new API.

I think we should also move away from `Foundation.Data` and replace it everywhere with `ByteBuffer`. It is way easier and efficient for all Metadata Encoders and Decoders to work directly with a ByteBuffer instead of `Foundation.Data` because they need to read/write binary data in network byte order. I think with this new API it is very unlike that anyone has to work with `Payload` directly. It will be just an implementation detail and can therefore use SwiftNIO types.

### Tests
Tests have not been added because almost all encoders and decoders are not implemented yet. I would like to introduce tests when we have actually implemented them.

### CI
~~This API uses result builders which were introduces in Swift 5.4 and Xcode 12.5. For Xcode 12.5 we also need macOS 11 which is not yet publicly available on GitHub Actions. I have requested access to the macOS 11 preview on GitHub for the RSocket GitHub Organization. However, it should compile and run fine on Linux! I would propose that we may want to disable macOS for some time. If that takes too long, we can also conditionally compile every API which uses result builders but that probably complicates things.~~
Update: we have access to macOS 11 and this PR now updates CI to use macOS 11 and Xcode 12.5.


Sources/RSocketCore/Extensions/Coder/CoderBuilder.swift
```diff
@@ -0,0 +1,124 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@resultBuilder
+public enum CoderBuilder: DecoderBuilderProtocol, EncoderBuilderProtocol {
```
nkristek, line 18: Could we also implement the `buildIf` and `buildEither` methods? This would make it possible to use control flow statements in the result builder closure.

Sources/RSocketCore/Extensions/Coder/Decoder.swift
```diff
@@ -0,0 +1,273 @@
...
+    public func decode(
+        _ payload: Payload, 
+        mimeType: ConnectionMIMEType
+    ) throws -> (Data?, Data) {
+        (payload.metadata, payload.data)
+    }
+}
+
+public struct AnyDecoder<Metadata, Data>: DecoderProtocol {
+    var _decoderBox: _AnyDecoderBase<Metadata, Data>
```
nkristek, line 39: Can this be private?
```suggestion
    private let _decoderBox: _AnyDecoderBase<Metadata, Data>
```

Sources/RSocketCore/Extensions/Coder/Decoder.swift
```diff
@@ -0,0 +1,273 @@
...
+    init<Decoder>(
```
nkristek, line 40: Should it be internal?
```suggestion
    internal init<Decoder>(
```

Sources/RSocketCore/Extensions/Coder/Decoder.swift
```diff
@@ -0,0 +1,273 @@
...
+        mimeType: ConnectionMIMEType
+    ) throws -> (Metadata, Data) {
+        if !isKnownUniquelyReferenced(&_decoderBox) {
+            _decoderBox = _decoderBox.copy()
+        }
+        return try _decoderBox.decode(payload, mimeType: mimeType)
+    }
+}
+
+class _AnyDecoderBase<Metadata, Data>: DecoderProtocol {
```
nkristek, line 56: Can be private (fileprivate in this case):
```suggestion
private class _AnyDecoderBase<Metadata, Data>: DecoderProtocol {
```

Sources/RSocketCore/Extensions/Coder/Decoder.swift
```diff
@@ -0,0 +1,273 @@
...
+        mimeType: ConnectionMIMEType
+    ) throws -> (Metadata, Data) {
+        fatalError("\(#function) in \(Self.self) is an abstract method and needs to be overridden")
+    }
+    func copy() -> _AnyDecoderBase<Metadata, Data> {
+        fatalError("\(#function) in \(Self.self) is an abstract method and needs to be overridden")
+    }
+}
+
+final class _AnyDecoderBox<Decoder: DecoderProtocol>: _AnyDecoderBase<Decoder.Metadata, Decoder.Data> {
```
nkristek, line 68:
```suggestion
private final class _AnyDecoderBox<Decoder: DecoderProtocol>: _AnyDecoderBase<Decoder.Metadata, Decoder.Data> {
```

nkristek (changes requested):
I like the idea of this PR and it will make it much easier for clients to create type safe environments.
Unfortunately, this PR has too many structs/classes in each file and they should be split out more. It is totally fine to have 2-3 types in the same file, but if there are more, it gets messy quite quickly.
Often times I noticed, that properties, that could be `private`, are implicitly `internal`, because there is no access control keyword. If they should be `internal` I would like it being explicitly set.

Sources/RSocketCore/Extensions/Authentication.swift
```diff
@@ -0,0 +1,143 @@
...
+
+public struct AuthenticationType: RawRepresentable, Hashable {
+    public let rawValue: String
+    public init(rawValue: String) {
+        self.rawValue = rawValue
+    }
+}
+
+public struct WellKnownAuthenticationTypeCode: RawRepresentable, Hashable {
+    /// rawValue is guaranteed to be between 0 and 127.
```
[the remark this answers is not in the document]
dnadoba: Inclusive. I have update the documentation to be more explicit about it.

Sources/RSocketCore/Extensions/Coder/CoderBuilder.swift
dnadoba: Yes, I have added them. I have also added `buildLimitedAvailability(_:)`.

dnadoba:
I have split the implementation into many smaller files as @nkristek suggested to simplify code review:
Screen Shot 2021-07-09 at 12 11 43 PM
Access control is now explicit and I have added @inlinable/@usableFromInline to almost all generic symbols to allow the compiler to create specializations.

Sources/RSocketCore/Extensions/Coder/ErasableMetadata.swift
```diff
@@ -0,0 +1,203 @@
+/*
```
nkristek, line 1: In this file we could split the extension of the encoders and decoders into separate files, wdyt?

Sources/RSocketCore/RequestDescription.swift
```diff
@@ -0,0 +1,111 @@
...
+    ) where 
+    Encoder: MetadataEncoder, 
+    Encoder.Metadata == Metadata
+    {
+        encoder = makeEncoder().eraseToAnyMetadataEncoder()
+    }
+}
+
+public struct FireAndForget<Request> {
+    public let encoder: AnyEncoder<Void, Request>
```
nkristek, line 41: Do we want to add a "normal" init to the other structs as well similar to `MetadataPush`?

nkristek (approved):
Good job on the PR!

Sources/RSocketCore/RequestDescription.swift
dnadoba: I would say we should not at first. `MetadataPush` has a normal `init(using:)` because we do not have a `MetadataEncoderBuilder`, only a `CompositeMetadataDecoderBuilder`. Maybe we should instead add a MetadataDecoder to be more in line with the over initializer, even if it does not add any other value compared to the other Builders.

OlegDokuka (approved):
LGTM

PR #54, opened 2021-07-19
fix custom debugDescription/description overload in RSocketTestUtilities

dnadoba:
### Motivation:
The Swift Compiler warns us that Payload conforms to `CustomDebugStringConvertible` twice, which is not supported by Swift.

### Modifications:
Conform `Payload` to `CustomStringConvertible` in `RSocketCore` and only to `CustomDebugStringConvertible` in `RSocketTestUtilities`. 

### Result:
Payload tries to read it's metadata and data as a UTF-8 string when `RSocketTestUtilities` is imported.
Otherwise it just prints the size of metadata/data in bytes.

### DIFF
--- a/Sources/RSocketCore/Payload/Payload.swift
+++ b/Sources/RSocketCore/Payload/Payload.swift
@@ -37,8 +37,9 @@ public struct Payload: Hashable {
     }
 }
 
-extension Payload: CustomDebugStringConvertible {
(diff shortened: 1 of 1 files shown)


nkristek (approved):

OlegDokuka (approved):
LGTM

PR #55, opened 2021-07-19
Routing Metadata Extension

dnadoba:
> Note: This PR builds on top of #53 

Add support for the [Routing Metadata Extension](https://github.com/rsocket/rsocket/blob/master/Extensions/Routing.md)

### Motivation:
We want to support easy Routing Metadata Encoding and Decoding

### Modifications:

- implement Routing Encoder & Decoder
- add new type `RoutingMetadata` to make it more convenient to create a route with only one tag 

### Result:
We can now use the `RoutingEncoder` and `RoutingDecoder` types. In addition, encoding a route with only a single tag can be written as:
```swift
Encoder()
    .encodeStaticMetadata("stock.price", .routing)
```
instead of wrapping it in an array literal.
```swift
Encoder()
    .encodeStaticMetadata(["stock.price"], .routing)
```
Note that the old syntax is still valid.


Sources/RSocketCore/Utility/ByteBuffer+LengthPrefix.swift
```diff
@@ -0,0 +1,55 @@
...
+    }
+}
+
+extension ByteBuffer {
+    /// Prefixes a message written by `writeMessage` with the number of bytes written as an `Integer`.
+    /// - Throws: If the number of bytes written during `writeMessage` can not be exactly represented as the given `Integer` i.e. if the number of bytes written is greater than `Integer.max`
+    /// - Returns: Number of total bytes written
+    @discardableResult
+    @inlinable
+    internal mutating func writeLengthPrefix<Integer: FixedWidthInteger>(
```
nkristek, line 32: Nice extension!

nkristek (approved):

OlegDokuka (approved):
LGTM

PR #56, opened 2021-07-21
Implement MIMETypeDecoder/Encoder

dnadoba:
> Note: depends on #55 and #53 

implement MIMETypeDecoder/Encoder

### Motivation:

Laying the ground work for CompositeMetadata by supporting encoding/deocding Well Known MIME Types and normal MIME Types.

### Modifications:

- add Decoder/Encoder
- add Tests

### Result:

We support encoding/decoding Well Known MIME Types and normal MIME Types


OlegDokuka (approved):
LGTM

Sources/RSocketCore/Extensions/MIMETypeDecoder.swift
```diff
@@ -0,0 +1,61 @@
...
+    ) {
+        self.wellKnownMimeTypes = wellKnownMimeTypes
+    }
+    
+    @inlinable
+    public func decode(from buffer: inout ByteBuffer) throws -> MIMEType {
+        guard let idOrLength = buffer.readInteger(as: Int8.self) else {
+            throw Error.invalid(message: "could not read MIME Type id/length")
+        }
+        if idOrLength >= 0 {
```
nkristek, line 38: Could we add a very small comment here explaining that all well known types start with a `1` and are thus negative?

nkristek (approved):

PR #61, opened 2021-08-18
Requester API with Decoder/Encoder

dnadoba:
This PR makes it possible to use the new Decoder/Encoder Pipeline in the ReactiveSwift Requester API.

### Motivation:

Make it easy to execute strongly typed requests.  

### Modifications:

- add `RequesterRSocket` protocol and conform RequesterAdapter to it:
```swift
public protocol RequesterRSocket {
    func execute<Metadata>(_ metadataPush: MetadataPush<Metadata>, metadata: Metadata) throws

    func execute<Request>(_ fireAndForget: FireAndForget<Request>, request: Request) throws

    func build<Request, Response>(
        _ requestResponse: RequestResponse<Request, Response>,
        request: Request
    ) -> SignalProducer<Response, Swift.Error>

    func build<Request, Response>(
        _ requestStream: RequestStream<Request, Response>,
        request: Request
    ) -> SignalProducer<Response, Swift.Error>

    func build<Request, Response>(
        _ requestChannel: RequestChannel<Request, Response>,
        initialRequest: Request,
        producer: SignalProducer<Request, Swift.Error>?
    ) -> SignalProducer<Response, Swift.Error>
}
```
- remove old request API because the new can do everything the old could.
- update examples and tests to use the new API

### Result:

We can now easily make request which use RSocket extensions e.g. CompositeMetadata or Routing.

#### method names build & execute 
The new requester methods for `MetadataPush` and `FireAndForget` are called `execute` because they start the work immediately. The methods for `RequestResponse`, `RequestStream` and `RequestChannel` are called `build` because they only start the work after the returned signal producer is started. This inconstancy is due to the limitation that we can't observe the progress of `FireAndForget` or `MetadataPush` streams. We may want to change that in the future so we can at least observe if the frame is flushed to the operating system. However, we can never say for sure that a `FireAndForget` or `MetadataPush` frame is actually received. This is by design of the protocol.

One could argue that this can be easily seen by the return types of those methods. The `execute` methods does not return anything and the `build` methods returns `SignalProducer`s, also know as a "cold signal". We would like to hear other opinions on that. If this is clear without the method name we could use just one method name or drop it all together by using `callAsFunction`.

#### callAsFunction
The first prototype used `callAsFunction` instead of `build`/`execute`. This makes the call site a bit shorter because we can drop the method name and use the `Requester` as if it is a function. After a discussion with @nkristek we decided that we want to give them a name initially because of the inconsistent start behavior between  `MetadataPush`/`FireAndForget` and `RequestResponse`/`RequestStream`/`RequestChannel`. Otherwise I think it would be nice to use `callAsFunction` here.

### DIFF
--- a/Sources/Examples/TimerClient/main.swift
+++ b/Sources/Examples/TimerClient/main.swift
@@ -6,14 +6,6 @@ import RSocketNIOChannel
 import RSocketReactiveSwift
 import RSocketWSTransport
 
-func route(_ route: String) -> Data {
-    let encodedRoute = Data(route.utf8)
-    precondition(encodedRoute.count <= Int(UInt8.max), "route is to long to be encoded")
-    let encodedRouteLength = Data([UInt8(encodedRoute.count)])
-
-    return encodedRouteLength + encodedRoute
-}
-
 extension URL: ExpressibleByArgument {
     public init?(argument: String) {
         guard let url = URL(string: argument) else { return nil }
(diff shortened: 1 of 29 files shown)


nkristek (approved):

OlegDokuka (approved):
Very nice!

PR #62, opened 2021-08-24
Remove Dependency On `NIO` Module

dnadoba:
Remove dependency on `NIO` and replace it with dependencies on `NIOCore`, `NIOPosix` or `NIOEmbedded`.

### Motivation:
[SwiftNIO has split its implementation from one module (`NIO`) into multiple smaller modules `NIOCore`, `NIOPosix` and `NIOEmbedded`](https://github.com/apple/swift-nio/releases/tag/2.32.0). We should no longer import `NIO` and instead use the smaller modules. This will result in binary size improvements, especially for iOS/macOS applications because they do not need `NIOPosix` and `NIOEmbedded`.

### Modifications:
- replace all target dependencies of `NIO` with `NIOCore`, `NIOPosix` or `NIOEmbedded` in `Package.swift`
- replace all imports of `NIO` with `NIOCore`, `NIOPosix` or `NIOEmbedded`

### Result:
Dependencies are now more fine grained.

Note: `NIOHTTP1`, which we use in `RSocketWSTransport` to support WebSocket as transport, still imports `NIO` and therefore `NIOPosix` and `NIOEmbedded`. This will probably only change in SwiftNIO 3.x because it is an [API breaking change](https://forums.swift.org/t/semantic-versioning-should-removing-a-dependency-be-a-semver-major/51179).


nkristek (approved):

OlegDokuka (approved):
LGTM

PR #63, opened 2021-08-25
Use ByteBuffer throughout RSocket

dnadoba:
Replace all occurrences of `Foundation.Data` with `NIOCore.ByteBuffer`.

### Motivation:
`ByteBuffer` is much more efficient than `Data`. We also convert a lot between `Data` and` ByteBuffer` and in very inefficient ways (i.e. by coping). We should remove them all and delay conversation to `Data` to the very last moment or don't do one at all.

### Modifications:

- replace all occurrences of`Foundation.Data` with `NIOCore.ByteBuffer`
- remove `import Foundation` if possible
- add new operator `asData()` to `DecoderProtocol` and `EncoderProtocol` to make it convenience to convert from/to `Foundation.Data`. This makes it also possible to convert to `Data` without making a copy by configuring the `byteTransferStrategy` to `.noCopy`.

### Result:
More efficient and simpler implementation.

### Future Directions
That was our biggest dependency on Foundation. We still use Foundation in a some of places: 
- in `KeepaliveHandler` to use the `TimeInterval` typealias to `Double`
- in `JSONDataDecoder/Encoder` to use `JSONDecoder/Encoder`
-  and in the new `asData()` operator

Replacing `TimeInterval` with `Double` is trivial. The over two are a bit more challenging. I think we would need to create a separate module e.g. `RSocketFoundationCompat` like `SwiftNIO` does with its `NIOFoundationCompat` module. 
This could be combined in the future with [cross-import overlays](https://gist.github.com/beccadax/8f35216340e829efa5e9c1fc090fda09) which would mean that `RSocketFoundationCompat` is implicitly imported if `RSocketCore` and `Foundation` are imported. AFAIK, this is currently behind a feature flag but already used but Apple.


Sources/Examples/VanillaClient/main.swift
```diff
@@ -23,7 +24,7 @@ struct VanillaClientExample: ParsableCommand {
         let client = try bootstrap.connect(to: .init(host: host, port: port)).first()!.get()
 
         let streamProducer = client.requester.requestStream(payload: .empty)
-        let requestProducer = client.requester.requestResponse(payload: Payload(data: Data("HelloWorld".utf8)))
+        let requestProducer = client.requester.requestResponse(payload: Payload(data: ByteBuffer(bytes: "HelloWorld".utf8)))
```
nkristek, line 27: Although this may be a tiny bit more performant, I think this API may be quite cumbersome to use. Can we add convenience extensions for `Data`? Through inlining it should be fine performance wise.

dnadoba: The Requester API which takes `Payload`s directly will be removed with this PR: https://github.com/rsocket/rsocket-swift/pull/61
The new API uses the Decoder/Encoder pipeline. That now has a convenience extension. I think we therefore do not need an additional extension on `Payload`.

dnadoba: @nkristek #61 is now merged and we no longer use `Data` at all in the examples. `ByteBuffer` already has spezial initializers for `String` and `String` also has a spezial initializer for `ByteBuffer`.
Swift Protobuf uses Data as it's input/output format. Support for protobuf should be implemented as extension and use the `ByteTransferStrategy.noCopy` to get `Data` from a `ByteBuffer` as an implementation detail. 

I think no user of RSocket Swift should actually use `Data` directly. If they do so, we should encourage them to implement their own extensions on `DecoderProtocol`/`EncoderProtocol`. 

What are the use cases you have in mind that require an extension on `Payload`?

nkristek (commented):
Why does this PR add a `pcap` file?

Sources/Examples/VanillaClient/main.swift
```diff
@@ -24,9 +25,9 @@ struct VanillaClientExample: ParsableCommand {
 
         let streamProducer = client.requester.build(
             RequestStream(),
-            request: Data()
+            request: ByteBuffer()
         )
-        let requestProducer = client.requester.build(RequestResponse(), request: Data("HelloWorld".utf8))
+        let requestProducer = client.requester.build(RequestResponse(), request: ByteBuffer(string: "HelloWorld"))
```
nkristek, line 30: A common case may be that the request payload is JSON encoded and the existing `JSONEncoder` API returns `Data`. Thus it might make it easier to use, if we add an overload that accepts `Data` as well.

Sources/Examples/VanillaClient/main.swift
nkristek: My reply is in this comment: https://github.com/rsocket/rsocket-swift/pull/63#discussion_r695807829

OlegDokuka (approved):

PR #71, opened 2022-12-20
Ayush yadav001/invalid header listener

ayush-yadav001:
Handling Invalid header error listener

**Motivation:**
Once we try initiate a connection and pass invalid header, In this case we are not receiving error call back 

**Modifications:**
Added resultHandler: @escaping (Result<Void, Swift.Error>) -> EventLoopFuture<Void>)-> EventLoopFuture<Void> in TransportChannelHandler protocol

**Result:**
Now if we pass invalid header  while connection ,will receive the error in result handler.

### DIFF
--- a/Sources/RSocketNIOChannel/ClientBootstrap.swift
+++ b/Sources/RSocketNIOChannel/ClientBootstrap.swift
@@ -88,6 +88,7 @@ extension ClientBootstrap: RSocketCore.ClientBootstrap {
                 }
             }
             .connect(host: endpoint.host, port: endpoint.port)
+        connectFuture.cascadeFailure(to: requesterPromise)
         return connectFuture.flatMap { channel in
             requesterPromise.futureResult.map { socket in
                 // initializing core client using channel object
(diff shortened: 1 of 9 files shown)


Sources/RSocketNIOChannel/ClientBootstrap.swift
```diff
@@ -65,15 +65,26 @@ extension ClientBootstrap: RSocketCore.ClientBootstrap {
...
+                            if case .failure(let error) = result {
+                                requesterPromise.fail(error)
+                                return requesterPromise.futureResult.eventLoop.makeFailedFuture(error)
+                            }
+                            return channel.pipeline.addRSocketClientHandlers(
+                                config: config,
+                                setupPayload: payload,
+                                responder: responder,
+                                connectedPromise: requesterPromise
+                            )
```
nkristek, line 86: Will the `RSocketClientHandlers` be added multiple times when this runs additionally to `upgradeComplete`? (as the handlers are already added there)

ayush-yadav001: No, i added the` RSocketClientHandlers`  in resultHandler which add only when we call ResultHandler() from `Sources/RSocketWSTransport/HTTPInitialRequestHandler.swift` file  when error occurs in `errorCaught` method, since in thats case `upgradeComplete` will not get called.

nkristek (approved):

OlegDokuka (approved):

PR #75, opened 2025-04-01
Xcode 16 Compatible changes 

ayush-yadav001:
changes in  Sources/RSocketCore/Extensions/RequestExamples.swift file which is not compiling with 
### Motivation:
Rsocket-Swift lib is not compiling with Xcode 16  
_[Explain here the context, and why you're making that change. What is the problem you're trying to solve.]_
Sources/RSocketCore/Extensions/RequestExamples.swift is not compiling with xcode 16 and has done the changes according to suggestion provided by Xcode16
### Modifications:
done the changes according to suggestion provided by Xcode16
_[Describe the modifications you've done.]_

### Result:
is compiling with Xcode16
_[After your change, what will change.]_

### DIFF
--- a/Sources/RSocketCore/Extensions/RequestExamples.swift
+++ b/Sources/RSocketCore/Extensions/RequestExamples.swift
@@ -15,8 +15,8 @@
  */
 
 /**
- This file contains examples of the `Coder`, `Encoder` and `Decoder` API. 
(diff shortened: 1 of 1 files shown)


ayush-yadav001:
forgot to add Sign off raising new PR

